PR #244, opened 2020-06-30
Wait for upstream job to be finished

kaweue:


kaweue:
recheck

test/env/run_e2e_tests.sh
```diff
@@ -6,6 +6,19 @@ BUILD_SCM_BRANCH=${BUILD_SCM_BRANCH:-"master"}
 BUILD_SCM_REVISION=${BUILD_SCM_REVISION:-"latest"}
 SKIP_TEST=${SKIP_TEST:-no}
 
+if [[ $BUILD_SCM_REVISION != "latest" ]] ; then
+    while : ; do
+        check_suites=$(curl -s -H "Accept: application/vnd.github.antiope-preview+json" https://api.github.com/repos/Juniper/contrail-operator/commits/$BUILD_SCM_REVISION/check-suites)
+        status=$(echo "$check_suites" | jq -cr '.check_suites[]|select(.app.slug == "google-cloud-build").status')
+        [[ $status != "completed" ]] && echo "Waiting for upstream job. Current status: $status" && sleep 120 && continue
```
Dysproz, line 13: I'm not a fan of sleep - especially for 2 minutes in CI job that may have timeout, because it adds fixed 2 minutes to job's time.
Wouldn't it be possible to make it somehow use loop which has smaller value of sleep (e.g. 5 seconds) and periodically check for condition?

Maybe I'm not into what's going on here exactly and why, but this sleep stresses me :)

katrybacka: I agree that we can try to poll this status more frequently :)

test/env/run_e2e_tests.sh
```diff
@@ -6,6 +6,19 @@ BUILD_SCM_BRANCH=${BUILD_SCM_BRANCH:-"master"}
...
+        conclusion=$(echo "$check_suites" | jq -cr '.check_suites[]|select(.app.slug == "google-cloud-build").conclusion')
+        [[ $conclusion == "success" ]] && break
+        echo "Upstream job failed with conclusion: $conclusion"
```
katrybacka, line 16: One more thing, at the first sight I agreed that e2e tests don't have to be run when the upstream job failed but are we sure that we don't want to know the status of e2e when for example only unit tests failed in GCB job? I know that it is possible to have a problem with phishing images but I'm not sure if we should stop e2e from running in case of failure od GCB job. I would like to know your opinion about that.

test/env/run_e2e_tests.sh
kaweue: Yeah I unfortunately this 2 minutes is not a coincident. As an unauthorised user I we issue 60 requests per hour https://developer.github.com/v3/#rate-limiting.

I may try to enable authentication if you would like to

kaweue: I will try to enable this authentication

Dysproz: Great :D I know that storing credentials for authentication in CI is always a big challenge, so if authentication would generate security problems then I may accept these 2 minutes ;)

kaweue: Fixed I moved this code to Jenkinsfile and used token and I reduced sleep to 5

kaweue: I will apply this Jenkinsfile on Jenkins as soon as you approve

kaweue:
recheck

PR #245, opened 2020-06-30
Add -f flag to the copy command in vroutercni container to force overwrite

psykulsk:
When contrail version is changed, or the node is restarted and pods are rescheduled again, vroutercni init container has to force overwrite the files created previously. Otherwise, cp command may fail with error:
cp: cannot create regular file '/host/opt_cni_bin/contrail-k8s-cni': Text file busy


psykulsk:
recheck

PR #255, opened 2020-07-07
Multus contrail configs support

Dysproz:
For Openshift in order to make Multus work with Contrail there has to be contrail config file in three locations:
- /etc/cni/net.d
- /etc/kubernetes/cni/net.d
- /var/run/multus/cni/net.d

Because bare kubernetes requires config only in /etc/cni/net.d then I removed optional path for config files from cluster info and created alternative container which in openshift dpeloyment will be added to vrouter pod and will copy contrail config to /etc/kubernees/cni/net.d and /var/run/multus/cni/net.d while /etc/cni/net.d will be handled as is used to be in bare kubernetes setup.

### DIFF
--- a/pkg/controller/vrouter/vrouter_controller.go
+++ b/pkg/controller/vrouter/vrouter_controller.go
@@ -469,8 +469,6 @@ func (r *ReconcileVrouter) Reconcile(request reconcile.Request) (reconcile.Resul
 			(&daemonSet.Spec.Template.Spec.InitContainers[idx]).Command = instanceContainer.Command
 		}
 		if container.Name == "vrouterkernelinit" {
-			reqLogger.Info("I detected vrouterkernelinit")
-
 			(&daemonSet.Spec.Template.Spec.InitContainers[idx]).EnvFrom = []corev1.EnvFromSource{{
 				ConfigMapRef: &corev1.ConfigMapEnvSource{
 					LocalObjectReference: corev1.LocalObjectReference{
@@ -483,8 +481,6 @@ func (r *ReconcileVrouter) Reconcile(request reconcile.Request) (reconcile.Resul
 			}
 		}
 		if container.Name == "vroutercni" {
-			reqLogger.Info("I detected vroutercni")
-
 			// vroutercni container command is based on the entrypoint.sh script in the contrail-kubernetes-cni-init container
 			command := []string{"sh", "-c",
 				"mkdir -p /host/etc_cni/net.d && " +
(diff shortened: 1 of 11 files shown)


pkg/k8s/cluster_info_test.go
```diff
@@ -96,8 +96,8 @@ func (suite *ClusterInfoSuite) TestCNIBinariesDirectory() {
 	suite.Assert().Equal(suite.VrouterClusterInfo.CNIBinariesDirectory(), "/opt/cni/bin", "Path should be /opt/cni/bin")
 }
 
-func (suite *ClusterInfoSuite) TestCNIConfigFilesDirectory() {
-	suite.Assert().Equal(suite.VrouterClusterInfo.CNIConfigFilesDirectory(), "/etc/cni", "Path should be /etc/cni")
+func (suite *ClusterInfoSuite) TestDeploymentType() {
+	suite.Assert().Equal(suite.VrouterClusterInfo.DeploymentType(), "k8s", "Path should be k8si")
```
mfigurski, line 100:
```suggestion
	suite.Assert().Equal(suite.VrouterClusterInfo.DeploymentType(), "k8s", "Path should be k8s")
```

Dysproz:
recheck

Dysproz:
recheck

Dysproz:
recheck ha

pkg/controller/vrouter/vrouter_controller.go
```diff
@@ -469,6 +469,8 @@ func (r *ReconcileVrouter) Reconcile(request reconcile.Request) (reconcile.Resul
 			(&daemonSet.Spec.Template.Spec.InitContainers[idx]).Command = instanceContainer.Command
 		}
 		if container.Name == "vrouterkernelinit" {
+			reqLogger.Info("I detected vrouterkernelinit")
```
psykulsk, line 472: I would either delete these log entries, or make them more descriptive, e.g.: "Adding vrouterkernelinit to init containers in daemonset pod spec"

psykulsk: And if you plan to leave it, then maybe change from Info to Debug

pkg/controller/vrouter/vrouter_controller.go
```diff
@@ -481,6 +483,8 @@ func (r *ReconcileVrouter) Reconcile(request reconcile.Request) (reconcile.Resul
 			}
 		}
 		if container.Name == "vroutercni" {
+			reqLogger.Info("I detected vroutercni")
```
psykulsk, line 486: Same commet as in line 472

pkg/controller/vrouter/vrouter_controller.go
Dysproz: Done, sorry

pkg/controller/vrouter/vrouter_controller.go
Dysproz: Done

PR #256, opened 2020-07-08
Add clusterIP to Certificate

katrybacka:


pkg/certificates/certificate_subject.go
```diff
@@ -29,3 +36,45 @@ func (c certificateSubjects) createCertificateSubjects() []certificateSubject {
 	return subjects
 
 }
+
+func (c certificateSubject) generateCertificateTemplate(alternativeIP string) (x509.Certificate, *rsa.PrivateKey, error) {
```
mfigurski, line 40: I would leave this as an function, not method. You just use subjectIP and don't change anything in subject struct.

katrybacka: Why? I did not change subject struct but I don't see a point in having this as a function not a method. Creation of certificate is for subject and it uses it inners

pkg/certificates/certificate_subject.go
```diff
@@ -1,7 +1,14 @@
...
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"fmt"
 	core "k8s.io/api/core/v1"
+	"net"
+	"time"
```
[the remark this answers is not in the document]
katrybacka: dooone

pkg/certificates/cacertificate.go
```diff
@@ -96,7 +96,7 @@ func caCertExistsInSecret(secret *corev1.Secret) bool {
 	return certOk && privKeyOk
 }
 
-func generateCaCerttificate() ([]byte, []byte, error) {
+func generateCaCertificate() ([]byte, []byte, error) {
 	caCertTemplate, caPrivKey, err := generateCaCerttificateTemplate()
```
shifali19, line 100: I know its not your mistake.But if you are at it, you can correct this spelling also.

PR #260, opened 2020-07-08
add clusterIP service for keystone

shifali19:


pkg/controller/keystone/keystone_controller.go
```diff
@@ -120,12 +120,26 @@ func (r *ReconcileKeystone) Reconcile(request reconcile.Request) (reconcile.Resu
...
+	keystoneService := newKeystoneService(keystone)
+	_, err := controllerutil.CreateOrUpdate(context.Background(), r.client, keystoneService, func() error { return nil })
+
 	keystonePods, err := r.listKeystonePods(keystone.Name)
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to list command pods: %v", err)
 	}
 
-	if err := r.ensureCertificatesExist(keystone, keystonePods); err != nil {
+	keystoneServiceList := &core.ServiceList{}
```
katrybacka, line 131: I would extract it to a separate method moreover I'm thinking that maybe we should watch on changes on that service (I know that in docs it was said that ClusterIP won't change but what about situation when someone will delete svc by accident or not)

pkg/controller/keystone/keystone_controller.go
```diff
@@ -120,12 +120,26 @@ func (r *ReconcileKeystone) Reconcile(request reconcile.Request) (reconcile.Resu
...
+	labelSelector := labels.SelectorFromSet(map[string]string{"service": keystone.Name})
+	listOpts := client.ListOptions{LabelSelector: labelSelector}
+	if err := r.client.List(context.TODO(), keystoneServiceList, &listOpts); err != nil {
+		return reconcile.Result{}, err
+	}
+	var keystoneClusterIP string
+	if len(keystoneServiceList.Items) > 0 {
+		keystoneClusterIP = keystoneServiceList.Items[0].Spec.ClusterIP
```
katrybacka, line 139: moreover I think that we should avoid situation when `keystoneClusterIP` because in that case we will create a certificates without serviceIP and it won't be regenerated, so I would exit here if it is empty and wait for it

pkg/controller/keystone/keystone_controller.go
```diff
@@ -120,12 +120,26 @@ func (r *ReconcileKeystone) Reconcile(request reconcile.Request) (reconcile.Resu
 		return reconcile.Result{}, nil
 	}
 
+	keystoneService := newKeystoneService(keystone)
```
katrybacka, line 123: this and createOrUpdate could be in method createKeystoneService Reconcile method would look much much better and more readable, but that's a nit

pkg/controller/keystone/keystone_controller.go
```diff
@@ -550,3 +563,19 @@ func (r *ReconcileKeystone) listKeystonePods(keystoneName string) (*core.PodList
...
 }
+
+func newKeystoneService(cr *contrail.Keystone) *core.Service {
+	return &core.Service{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      cr.Name + "-service",
+			Namespace: cr.Namespace,
+			Labels:    map[string]string{"service": cr.Name},
+		},
+		Spec: core.ServiceSpec{
```
katrybacka, line 574: setting Spec should be inside the mutate function of CreateOrUpdate since it can be overwritten by Get method called inside CreateOrUpdate. In description of this method is said that desired state of object should be set by mutate function

pkg/apis/contrail/v1alpha1/keystone_types.go
```diff
@@ -25,9 +25,10 @@ type KeystoneConfiguration struct {
 // KeystoneStatus defines the observed state of Keystone
 // +k8s:openapi-gen=true
 type KeystoneStatus struct {
-	Active bool     `json:"active,omitempty"`
-	Port   int      `json:"port,omitempty"`
-	IPs    []string `json:"ips,omitempty"`
+	Active    bool     `json:"active,omitempty"`
+	Port      int      `json:"port,omitempty"`
+	IPs       []string `json:"ips,omitempty"`
+	ClusterIP string   `json:"clusterIP,omitempty"`
```
katrybacka, line 31: Ohhh, we forgot about provision manager

katrybacka: please could you change the IP of Keystone there?

shifali19:
recheck ha

shifali19:
recheck

shifali19:
recheck

pkg/controller/keystone/keystone_controller.go
```diff
@@ -550,3 +570,40 @@ func (r *ReconcileKeystone) listKeystonePods(keystoneName string) (*core.PodList
...
+	labelSelector := labels.SelectorFromSet(map[string]string{"service": keystoneName})
+	listOpts := client.ListOptions{LabelSelector: labelSelector}
+	if err := r.client.List(context.TODO(), keystoneServiceList, &listOpts); err != nil {
+		return &core.ServiceList{}, err
+	}
+	return keystoneServiceList, nil
+}
+
+func (r *ReconcileKeystone) ensureServiceExists(keystone *contrail.Keystone,
+) error {
```
tomaszmj, line 585: nit: I think this can be in one line

shifali19:
recheck ha

pkg/apis/contrail/v1alpha1/keystone_types.go
```diff
@@ -25,9 +25,10 @@ type KeystoneConfiguration struct {
 // KeystoneStatus defines the observed state of Keystone
 // +k8s:openapi-gen=true
 type KeystoneStatus struct {
-	Active bool     `json:"active,omitempty"`
-	Port   int      `json:"port,omitempty"`
-	IPs    []string `json:"ips,omitempty"`
+	Active    bool     `json:"active,omitempty"`
+	Port      int      `json:"port,omitempty"`
+	IPs       []string `json:"ips,omitempty"`
```
kaweue, line 30: Is this IPs used anywhere?

pkg/apis/contrail/v1alpha1/keystone_types.go
kaweue, line 31: Maybe let's name it `IPaddress` or just `IP`. ClusterIP name suggest that it is "ClusterIP" however it could be something different like LB IP or even something external

pkg/apis/contrail/v1alpha1/keystone_types.go
kaweue: If not let's delete this

pkg/apis/contrail/v1alpha1/keystone_types.go
kaweue: Let's consider unifying IP address and a Port to one field `Endpoint`

pkg/apis/contrail/v1alpha1/keystone_types.go
shifali19: Deleted

pkg/apis/contrail/v1alpha1/keystone_types.go
shifali19: NA

PR #262, opened 2020-07-09
Add rhel tag to selected images in in openshift manifest

Dysproz:
In this review I add to openshift manifest rhel tag to selected images (all from juniper hub expect for operator, statusmonitor and provisioner - because these are not present with rhel tag) as target version for final release.

Also I changed vrouter kernel init container from private repository to hub juniper image.

Also added information in readme about minimal version of openshift-install binary required.

### DIFF
--- a/deploy/openshift/README.md
+++ b/deploy/openshift/README.md
@@ -2,7 +2,7 @@
 Deployment depends strongly on Openshift installation which is described in this [documentation](https://docs.openshift.com/container-platform/4.1/installing/installing_aws/installing-aws-customizations.html)
 
 Prerequisities that have to be fulfilled in order to dpeloy Contrail with operator on Openshift:
-* openshift-install binary ([download](https://cloud.redhat.com/openshift/install))
+* openshift-install binary (>=4.4.8) ([download](https://cloud.redhat.com/openshift/install))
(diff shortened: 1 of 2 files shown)


deploy/openshift/README.md
```diff
@@ -2,7 +2,7 @@
 Deployment depends strongly on Openshift installation which is described in this [documentation](https://docs.openshift.com/container-platform/4.1/installing/installing_aws/installing-aws-customizations.html)
 
 Prerequisities that have to be fulfilled in order to dpeloy Contrail with operator on Openshift:
-* openshift-install binary ([download](https://cloud.redhat.com/openshift/install))
+* openshift-install binary (>=4.4.8) ([download](https://cloud.redhat.com/openshift/install))
```
psykulsk, line 5: Can you also add, in line 36 of README.md, information to not use the -rhel tags, just plain master.1245?
E.g. : 
Under *CONTRAIL_VERSION* field enter proper Contrail container build tag, available in the hub.juniper.net/contrail-nightly registry.
Do not specify the `-rhel` suffix in the build tag, it will be appended automatically where needed.

Dysproz: Done

PR #263, opened 2020-07-10
Implementation of Fernet keys rotation for Keystone

katrybacka:
This PR introduces Fernet keys rotation that are the base for creating and validating tokens by Keystone. This rotation should work like that:

- it should change staged key to primary key (primary key is the one with the highest index)
- primary becomes secondary (since after first step it does not have the highest index)
- new staged key needs to be generated as key with index "0"
- number of keys cannot exceed maxActiveKeys with is computed by controller using time of token expiration, allow token expired window and rotationInterval.

All of the mentioned fields have default values.


katrybacka:
recheck ha

katrybacka:
recheck ha

pkg/controller/keystone/keystone_controller.go
```diff
@@ -76,6 +75,14 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Watch for changes to secondary resource Postgres and requeue the owner Keystone
```
mfigurski, line 78: Postgres? 🙃

katrybacka: done

katrybacka:
recheck ha

katrybacka:
recheck

PR #265, opened 2020-07-10
Common status base structure for all services

kaweue:


pkg/apis/contrail/v1alpha1/status.go
```diff
@@ -0,0 +1,25 @@
...
+
+import appsv1 "k8s.io/api/apps/v1"
+
+// Status is the status of the service.
+// +k8s:openapi-gen=true
+type Status struct {
+	Active        bool  `json:"active,omitempty"`
+	Replicas      int32 `json:"replicas,omitempty"`
+	ReadyReplicas int32 `json:"readyReplicas,omitempty"`
+}
```
Dysproz, line 11: Is it going to be used in other resources apart from memcached?

kaweue: The plan is to use for all resources. We plan to adjust other resources one by one

Dysproz: Great :+1:

deploy/crds/contrail.juniper.net_memcacheds_crd.yaml
```diff
@@ -4,15 +4,21 @@ metadata:
...
   additionalPrinterColumns:
+  - JSONPath: .status.replicas
+    name: Replicas
+    type: integer
+  - JSONPath: .status.readyReplicas
+    name: Ready_Replicas
+    type: integer
+  - JSONPath: .status.endpoint
+    name: Endpoint
+    type: string
```
psykulsk, line 15: Nit: maybe, instead of two columns, there could be one with readyReplicas/replicas? Same style as is printed for core k8s resources, like Deployments, pods, etc.?
I think it can be done purely in the JSONPath (not 100% sure), like:

>  \- JSONPath: ".status.readyReplicas/.status.replicas"
>     name: Ready
>     type: string

pkg/apis/contrail/v1alpha1/status.go
```diff
@@ -0,0 +1,25 @@
+package v1alpha1
+
+import appsv1 "k8s.io/api/apps/v1"
+
+// Status is the status of the service.
+// +k8s:openapi-gen=true
+type Status struct {
+	Active        bool  `json:"active,omitempty"`
+	Replicas      int32 `json:"replicas,omitempty"`
```
shifali19, line 9: NIT: Since we have 2 replica fields, may be we could call this one "intendedReplicas" to avoid confusion.

deploy/crds/contrail.juniper.net_memcacheds_crd.yaml
kaweue: I will try that

kaweue: I think only "simple JSON Paths" are supported

pkg/apis/contrail/v1alpha1/status.go
kaweue: I would leave it as replicas

PR #275, opened 2020-07-16
Add script that generate encoded DOCKER_CONFIG for Openshift deployment

Dysproz:
This short script generates DOCKER_CONFIG which should be provided in config to install-manifests for Openshift

### DIFF
--- a/deploy/openshift/tools/docker-config-generate/generate-docker-config.sh
+++ b/deploy/openshift/tools/docker-config-generate/generate-docker-config.sh
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+if [[ "$OSTYPE" == "darwin"* ]]; then
+        BASE64_COMMAND='base64'
+else
+        BASE64_COMMAND='base64 -w 0'
+fi
+
 get_parameters() {
     read -p "Docker registry: "  REGISTRY
     read -p "Docker username: "  USERNAME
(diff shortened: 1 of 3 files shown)


deploy/openshift/tools/docker-config-generate/generate-docker-config.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
kaweue, line 1: Why not #!/usr/bin/env bash?

Dysproz: Done

psykulsk:
recheck

Dysproz:
recheck ha

Dysproz:
recheck ha

deploy/openshift/tools/docker-config-generate/generate-docker-config.sh
```diff
@@ -0,0 +1,19 @@
...
+    read -p "Docker registry: "  REGISTRY
+    read -p "Docker username: "  USERNAME
+    read -s -p "Docker secret: "  SECRET
+}
+
+generate_base64() {
+    AUTH=$(echo $USERNAME:$SECRET | base64 -w 0)
+    RENDERED_JSON=$(echo {\"auths\":{\"${REGISTRY}\":{\"username\": \"${USERNAME}\",\"password\":\"${SECRET}\",\"auth\":\"${AUTH}\"}}})
+    DOCKER_CONFIG=$(echo $RENDERED_JSON | base64 -w 0)
+}
```
psykulsk, line 13: Unfortunately base64 on macos does not have the -w option. And most of the people that are going to be using this, are using macos. I've seen suggestions to use base64 encoder built in openssl (e.g. `openssl base64`) that works fine on most OS's and doesn't add new lines.

psykulsk: So I would try changing `base64 -w 0` to `openssl base64` and adding a comment about that it is used for compatibility reasons.

Dysproz: Done

Dysproz:
recheck

PR #276, opened 2020-07-16
Upgrade keystone using DeleteFirst strategy

shifali19:


pkg/controller/keystone/keystone_controller.go
```diff
@@ -383,12 +332,8 @@ func newKeystoneSTS(cr *contrail.Keystone) *apps.StatefulSet {
 							Args:            []string{"-c", initDBScript},
 							Env: []core.EnvVar{
 								{
-									Name: "MY_POD_IP",
-									ValueFrom: &core.EnvVarSource{
-										FieldRef: &core.ObjectFieldSelector{
-											FieldPath: "status.podIP",
-										},
-									},
+									Name:  "MY_DB_IP",
```
kaweue, line 335: Let's name it maybe PSQL_ENDPOINT or PSQL_IP

pkg/controller/postgres/postgres_controller.go
```diff
@@ -110,12 +115,18 @@ func (r *ReconcilePostgres) Reconcile(request reconcile.Request) (reconcile.Resu
 		return reconcile.Result{}, nil
 	}
 
+	postgresClusterIP := "0.0.0.0"
```
kaweue, line 118: I think that this `postgresClusterIP` variable will have some value always from `postgresService.Spec.ClusterIP` (line 123)

pkg/controller/postgres/postgres_controller.go
```diff
@@ -367,3 +378,29 @@ func getCommand(containers []*contrail.Container, containerName string) []string
...
 	return c.Command
 }
+
+func (r *ReconcilePostgres) ensureServiceExists(postgres *contrail.Postgres) (*core.Service, error) {
+	postgresService := newPostgresService(postgres)
+	_, err := controllerutil.CreateOrUpdate(context.Background(), r.client, postgresService, func() error {
+		postgresService.Spec.Ports = []core.ServicePort{
+			{Port: 5432, Protocol: "TCP"},
+		}
+		postgresService.Spec.Selector = map[string]string{"app": postgres.Name}
```
kaweue, line 388: Maybe we could just copy label from `postgres` object. It will always match in that case

pkg/controller/postgres/postgres_controller.go
```diff
@@ -367,3 +378,29 @@ func getCommand(containers []*contrail.Container, containerName string) []string
...
+	err = controllerutil.SetControllerReference(postgres, postgresService, r.scheme)
+	return postgresService, err
+}
+
+func newPostgresService(cr *contrail.Postgres) *core.Service {
+	return &core.Service{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      cr.Name + "-service",
+			Namespace: cr.Namespace,
+			Labels:    map[string]string{"service": cr.Name},
```
kaweue, line 403: Let's use the same label as for cr object

test/e2e/ha/openstack_services_test.go
```diff
@@ -96,6 +115,15 @@ func TestHAOpenStackServices(t *testing.T) {
 			t.Run("then all services should have 3 ready replicas", func(t *testing.T) {
 				assertOpenStackReplicasReady(t, w, 3)
 			})
+
+			t.Run("then the keystone service should handle request for a token", func(t *testing.T) {
+				t.Skip()
```
kaweue, line 120: Why is this test skipped?

test/e2e/ha/openstack_services_test.go
```diff
@@ -164,11 +192,22 @@ func assertOpenStackReplicasReady(t *testing.T, w wait.Wait, r int32) {
 		t.Parallel()
 		assert.NoError(t, w.ForReadyDeployment("memcached-deployment", r))
 	})
+	t.Run(fmt.Sprintf("then a Keystone StatefulSet has %d ready replicas", r), func(t *testing.T) {
+		t.Skip()
```
kaweue, line 196: Why is this test skipped?

test/e2e/ha/openstack_services_test.go
```diff
@@ -185,6 +224,20 @@ func assertOpenStackPodsHaveUpdatedImages(t *testing.T, f *test.Framework, manag
 		}.ForPodImageChange(f.KubeClient, "Memcached="+manager.Spec.Services.Memcached.Name, mmContainerImage, "memcached")
 		assert.NoError(t, err)
 	})
+
+	t.Run("then keystone has updated image", func(t *testing.T) {
+		t.Skip()
```
kaweue, line 229: Why is this test skipped?

test/e2e/ha/openstack_services_test.go
```diff
@@ -96,6 +115,15 @@ func TestHAOpenStackServices(t *testing.T) {
 			t.Run("then all services should have 3 ready replicas", func(t *testing.T) {
 				assertOpenStackReplicasReady(t, w, 3)
 			})
+
+			t.Run("then the keystone service should handle request for a token", func(t *testing.T) {
+				t.Skip()
+				keystoneProxy := proxy.NewSecureClient("contrail", "keystone-keystone-statefulset-0", 5555)
```
kaweue, line 121: As we are introducing services. I think we should try to proxy traffic not to a pod but to service.

I think this might be it https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/

pkg/controller/keystone/keystone_controller.go
katrybacka: PSQL_IP pls since endpoint sounds like sth more than ip

pkg/controller/postgres/postgres_controller.go
katrybacka: you're right

test/e2e/ha/openstack_services_test.go
katrybacka: will we be able to access ClusterIP this way?

shifali19: We will use service name here

kaweue: We have to change the whole path like here https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#manually-constructing-apiserver-proxy-urls

PR #277, opened 2020-07-17
Remove StorageClass and PersistentVolumes namespaces and OwnerReferences to Custom Resources.

psykulsk:
The strange bug with zookeeper resources deletion (ticket: https://contrail-jws.atlassian.net/browse/CEM-16823) was actually not specific to the Openshift deployment. It happened on a Kind cluster as well when the kube-controller-manager restart was triggered manually. When this k8s component restarts, k8s Garbage Collection kicks in. 

Our zookeeper and cassandra Custom Resources are namespace-scoped resources. They are tied to the specific namespace they have been created in. The StorageClass and PersistentVolumes objects are cluster-scoped. They are not tied to any namespace. Zookeeper and Cassandra controllers created both the StorageClass "local-storage" and the PersistentVolumes. They also set the OwnerReference of these objects to the Zookeeper and Cassandra CRs respectively. This cannot be done. 
[https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/)
Part of the k8s GC documentation says:

> Cross-namespace owner references are disallowed by design. This means:
Namespace-scoped dependents can only specify owners in the same namespace, and owners that are cluster-scoped.
Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners.

In this PR I remove the lines that add the owner reference to PVs and the StorageClass. I also remove some obsolete Openshift files.

Having no OwnerReference has on downside, after the statefulsets of Zookeeper or Cassandra will be deleted, admin will have to remove the PVs manually. But in normal operation this should not happen. If the pods get deleted, containers killed, nodes get rebooted, the previously created PVCs will still be valid and new pods will use them. 

If we want to have a PVs that are automatically cleaned up and created, then we either have to implement our own controller for the StorageClass, go back to using the  default StorageClass or use an external storage provisioner like this one: https://github.com/rancher/local-path-provisioner


psykulsk:
recheck

Dysproz:
recheck

psykulsk:
recheck

kaweue:
This approach changes current behaviour. SC and PV are not deleted with cluster deletion. 
Consequence is that operator will create some resources and don't clean up them.
I'm not sure if this is what we expect from operator.

PR #288, opened 2020-07-28
Rabbitmq HA failover bug fix

Dysproz:
This PR fix bug where after node reboot, rabbitmq pod on rebooted node does not join cluster of other rabbitmqs.
I added cluster formation plugin which is configured with classic config to enable auto join to cluster (https://www.rabbitmq.com/cluster-formation.html#peer-discovery-classic-config).

Each rabbitmq has cluster nodes specified in config so when there's reboot it'll automatically join cluster.
Moreover, as described in documentation, to prevent situation where pod will not be able to join cluster because it'll have old data and during it's reboot some other changes happened to cluster I added to command run in pod `rabbitmqctl forget_cluster_node` which will ensure that pod will not try to join with old configuration and new session will be established.

I see that there's also an option to configure this plugin especially for k8s where peer discovery is done via k8s cluster, but it'd require some more changes in architecture of how we handle rabbitmq pods and since this bug is for R2008 then I'll just create improvement ticket for migration to k8s cluster peer discovery and this PR will be solution for R2008.

### DIFF
--- a/pkg/apis/contrail/v1alpha1/rabbitmq_types.go
+++ b/pkg/apis/contrail/v1alpha1/rabbitmq_types.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"sort"
 	"strconv"
-	"strings"
 
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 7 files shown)


Dysproz:
recheck ha

Dysproz:
recheck

Dysproz:
recheck

kaweue:
I would suggest to enable some e2e test that would verify that.

https://github.com/Juniper/contrail-operator/blob/R2008/test/e2e/ha/scale_core_contrail_services_test.go#L232

Dysproz:
recheck

Dysproz:
recheck ha

Dysproz:
recheck

Dysproz:
recheck ha

Dysproz:
recheck ha

PR #289, opened 2020-07-28
Additional parameters for keystone

michalskalski:
This change introduce new parameters for keystone resource:

  * authProtocol - 'http' or 'https' part of auth url which instructs
    clients to use encrypted connection or plain text
  * region - defines region used by keystone server and clients
  * projectDomain and userDomain - instructs clients which domain should
    be used

Those parameters are required when external keystone is in use to
properly configure clients in cluster. In that case keystone resource is
a source of information about external cluster and is not directly
managed by controler. It is external provisioner responsibility to mark
keyston resource active and set appropriate IP address.

Jira: CEM-17834


michalskalski:
recheck

michalskalski:
recheck ha

test/env/deploy/cluster.yaml
```diff
@@ -218,6 +218,12 @@ spec:
           memcachedInstance: "memcached"
           postgresInstance: postgres
           listenPort: 5555
+          authProtocol: https
```
kaweue, line 221: Does it need to be specified if we have some default values?

kaweue: I mean I think that at least one e2e should be executed with default values

pkg/controller/keystone/keystone_controller.go
```diff
@@ -238,6 +244,43 @@ func (r *ReconcileKeystone) Reconcile(request reconcile.Request) (reconcile.Resu
 	return reconcile.Result{}, r.updateStatus(keystone, sts, keystoneClusterIP)
 }
 
+func (r *ReconcileKeystone) setDefaultValues(k *contrail.Keystone) error {
+	patch := client.MergeFrom(k.DeepCopy())
```
kaweue, line 248: Maybe this patch could be moved closer to r.client.Patch

pkg/apis/contrail/v1alpha1/tests/config_test.go
```diff
@@ -1783,3 +1792,17 @@ AUTHN_DOMAIN = Default
 cafile = /etc/ssl/certs/kubernetes/ca-bundle.crt
 ;AUTHN_TOKEN_URL = http://127.0.0.1:35357/v2.0/tokens
 `
+var configKeystoneAuthConf = `[KEYSTONE]
+#memcache_servers=localhost:11211
```
kaweue, line 1796: Let's remove that commented code from template

contrail-provisioner/main.go
```diff
@@ -829,8 +829,15 @@ func databaseNodes(contrailClient *contrail.Client, nodeList []*types.DatabaseNo
 }
 
 func setupAuthKeystone(client *contrail.Client, keystoneAuthParameters *KeystoneAuthParameters) {
+	// CEM-17791 - AddEncryption expects http url
```
kaweue, line 832: This a public code so I'm not sure if we should expose our internal Jira tickets

pkg/apis/contrail/v1alpha1/templates/webui_config.go
```diff
@@ -16,7 +16,12 @@ config.endpoints = {};
 config.endpoints.apiServiceType = "ApiServer";
 config.endpoints.opServiceType = "OpServer";
 config.regions = {};
-config.regions.RegionOne = "https://{{ .KeystoneIP }}:{{ .KeystonePort }}/v3";
+{{- /* HA core contrail tests are run without keystone, which causes invalid config generation. */}}
```
kaweue, line 19: I mean not only tests are run without keystone but also some production deployments. I suggest a change in this description

test/env/deploy/cluster.yaml
michalskalski: Removed those defaults from this file and e2e tests files, left just one change of region to test this functionality on deployed environment.

pkg/controller/keystone/keystone_controller.go
michalskalski: Moved this part to manager. Should be done just once, and not checked constantly at runtime.

pkg/apis/contrail/v1alpha1/tests/config_test.go
michalskalski: Done

contrail-provisioner/main.go
michalskalski: Pointed to the commit which is already merged, not sure what version of the contrail-go-api we use, but if we could use the newest, then this addition wouldn't be needed.

pkg/apis/contrail/v1alpha1/templates/webui_config.go
michalskalski: Changed comment.

PR #292, opened 2020-07-30
Cassandra readiness probe

lrebeja:
Changed Cassandra readiness probe to check only one Cassandra instance running in a current pod.

Casandra cluster is active when at least the number of ready pods is more than half of desired.

### DIFF
--- a/pkg/apis/contrail/v1alpha1/cassandra_types.go
+++ b/pkg/apis/contrail/v1alpha1/cassandra_types.go
@@ -244,17 +244,13 @@ func (c *Cassandra) SetInstanceActive(client client.Client, activeStatus *bool, 
 		sts); err != nil {
 		return err
 	}
-	active := false
+	*activeStatus = false
 	var acceptableReadyReplicaCnt = *sts.Spec.Replicas/2 + 1
 	if sts.Status.ReadyReplicas >= acceptableReadyReplicaCnt {
-		active = true
+		*activeStatus = true
 	}
 
-	*activeStatus = active
-	if err := client.Status().Update(context.TODO(), c); err != nil {
-		return err
-	}
-	return nil
+	return client.Status().Update(context.TODO(), c)
 }
 
 // ManageNodeStatus manages the status of the Cassandra nodes.
--- a/pkg/apis/contrail/v1alpha1/cassandra_types.go
+++ b/pkg/apis/contrail/v1alpha1/cassandra_types.go
@@ -240,7 +240,21 @@ func (c *Cassandra) PodIPListAndIPMapFromInstance(instanceType string, request r
 
 // SetInstanceActive sets the Cassandra instance to active.
 func (c *Cassandra) SetInstanceActive(client client.Client, activeStatus *bool, sts *appsv1.StatefulSet, request reconcile.Request) error {
-	return SetInstanceActive(client, activeStatus, sts, request, c)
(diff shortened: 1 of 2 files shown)


pkg/apis/contrail/v1alpha1/cassandra_types.go
```diff
@@ -240,7 +240,21 @@ func (c *Cassandra) PodIPListAndIPMapFromInstance(instanceType string, request r
...
+		return err
+	}
+	active := false
+	var acceptableReadyReplicaCnt = *sts.Spec.Replicas/2 + 1
+	if sts.Status.ReadyReplicas >= acceptableReadyReplicaCnt {
+		active = true
+	}
+
+	*activeStatus = active
+	if err := client.Status().Update(context.TODO(), c); err != nil {
```
kaweue, line 254: I think that you can just return directly value from Update method.
'return client.Status().Update()

pkg/apis/contrail/v1alpha1/cassandra_types.go
```diff
@@ -240,7 +240,21 @@ func (c *Cassandra) PodIPListAndIPMapFromInstance(instanceType string, request r
 
 // SetInstanceActive sets the Cassandra instance to active.
 func (c *Cassandra) SetInstanceActive(client client.Client, activeStatus *bool, sts *appsv1.StatefulSet, request reconcile.Request) error {
-	return SetInstanceActive(client, activeStatus, sts, request, c)
+	if err := client.Get(context.TODO(), types.NamespacedName{Name: sts.Name, Namespace: request.Namespace},
+		sts); err != nil {
+		return err
+	}
+	active := false
```
kaweue, line 247: Could activeStatus variable be used directly here instead of having additional local variable?

pkg/controller/cassandra/sts.go
```diff
@@ -45,7 +45,7 @@ spec:
             command:
             - /bin/bash
             - -c
-            - "seeds=$(for i in $(ls /mydata/*.yaml); do echo $(basename $i .yaml); done) &&  for seed in $(echo $seeds); do if [[ $(nodetool status | grep $seed |awk '{print $1}') != 'UN' ]]; then exit -1; fi; done"
+            - "if [[ $(nodetool status | grep ${POD_IP} |awk '{print $1}') != 'UN' ]]; then exit -1; fi;"
```
kaweue, line 48: Maybe this command could be simplified command [[ returns already 0 or 1. And kubelet just check of status code is 0

pkg/apis/contrail/v1alpha1/cassandra_types.go
lrebeja: fixed

pkg/apis/contrail/v1alpha1/cassandra_types.go
lrebeja: fixed

pkg/controller/cassandra/sts.go
lrebeja: I prefer explicitly specify when it fails. It's more readable this way. Also, I tried to use expression inside [[]] and it didn't work. I guess I missed something. But my main point if you don't explicitly specifies the exit condition then it's not clear.

kaweue: ok

PR #297, opened 2020-08-03
CEM-15500 WebUI modify and extend Status Field in CR

katrybacka:
Add rediness probe to WebUI container

Skip WebUI related test cases in HA test suite because of being strong coupled with Keystone

Add readiness and liveness probe to WebUI's Redis

Add Service statuses to WebUI CR

Add new lines at the end of files

Fix pod statuses initialization

Change slice to map of ServiceStatuses


katrybacka:
recheck ha

PR #302, opened 2020-08-04
Incresed the timeout of the cassandra readiness probe to 15 seconds

psykulsk:
Context in the comment section https://contrail-jws.atlassian.net/browse/CEM-17977


psykulsk:
recheck

PR #310, opened 2020-08-11
Added all the modified files

vijayrap:


vijayrap:
vraparla-mbp:crds vraparla$ kubectl get contrailstatusmonitor -n contrail
NAME                      STATUS
cassandra1                Active
contrail-analytics-api    Functional
contrail-api              Functional
contrail-collector        Functional
contrail-device-manager   initializing
contrail-schema           Functional
contrail-svc-monitor      Functional
keystone                  Active
memcached                 Active
postgres                  Active
provmanager1              Active
rabbitmq1                 Active
redis                     Functional
webuijob                  Functional
webuiweb                  Functional
zookeeper1                Active
vraparla-mbp:crds vraparla$ 

--------------------

Ran couple of times.

deploy/crds/contrail.juniper.net_postgres_crd.yaml
```diff
@@ -9,7 +9,7 @@ spec:
     listKind: PostgresList
     plural: postgres
     singular: postgres
-  scope: Namespaced
```
kaweue, line 12: Please update operator-sdk to version v0.17.1 as described in documentation.

test/env/deploy/cluster.yaml
```diff
@@ -181,8 +181,8 @@ spec:
         name: command
       spec:
         commonConfiguration:
-          activate: true
-          create: true
+          activate: false
```
kaweue, line 184: Why is this changed from true to false?

pkg/controller/contrailmonitor/contrailmonitor_controller.go
```diff
@@ -0,0 +1,453 @@
...
+}
+
+// Reconcile reads that state of the cluster for a Contrailmonitor object and makes changes based on the state read
+// and what is in the Contrailmonitor.Spec
+// TODO(user): Modify this Reconcile function to implement your Controller logic.  This example creates
+// a Pod as an example
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileContrailmonitor) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
kaweue, line 144: I don't see any unit tests for that controller. As far as I know there is a requirement that a new code should be covered by unit tests

kaweue: Moreover I strongly suggest to extend one of the e2e tests that will verify happy path. Just a single simple check

pkg/controller/contrailmonitor/contrailmonitor_controller.go
```diff
@@ -0,0 +1,453 @@
...
+			return reconcile.Result{}, err
+		}
+
+	}
+
+	clist, err := r.getConfiglist()
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	fmt.Println(clist.Items[0].Status.ServiceStatus)
```
kaweue, line 322: It looks like a debug log. Please use logger or remove it

pkg/controller/contrailmonitor/contrailmonitor_controller.go
```diff
@@ -0,0 +1,453 @@
...
+			return reconcile.Result{}, err
+		}
+	}
+	provisionmanager, err := r.getProvisionmanager(instance)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	serIns3 := &contrailv1alpha1.Contrailstatusmonitor{ObjectMeta: metav1.ObjectMeta{Name: provisionmanager.Name, Namespace: "contrail"}}
+
+	if provisionmanager.Status.Active == nil {
```
kaweue, line 357: What if status.Active is false?

kaweue: Will we report "Active"?

deploy/crds/contrail.juniper.net_postgres_crd.yaml
vijayrap: As per  document, for 2008R  we are using the older version, I have noticed u have changes done in master not in 2008R.

pkg/controller/contrailmonitor/contrailmonitor_controller.go
vijayrap: I have not included unit test or e2e test cases here, I have noticed it, the reason is,  I get very inconstant results,  One day, I get one results, other day, i get completly diffrent results., I like you to suggest me what is wrong in the code that gives varied results.  for example,  even if the postgres service is active as it updates my additional print coloums crd instance it updates the field as Not active. it is very random

pkg/controller/contrailmonitor/contrailmonitor_controller.go
vijayrap: I will remove it.

deploy/crds/contrail.juniper.net_postgres_crd.yaml
kaweue: but this is PR to master

pkg/controller/contrailmonitor/contrailmonitor_controller.go
vijayrap: Yes, WE need to report Active/NotActive for postgres, provisionmanager, zookeeper, cassandra etc.

Yes, I have noticed  the point that you have raised, These fields in the crds are boolean pointers, I can check boolean pointer as  the GCP build raised Panic error,  which is not raised by the blazel build, so, after couple of iterations , I have decided with 2 values nil and true, I tried to even check the IsAlive function that cassandra, zookeper, rabbitmq used to to test the pod activity, I noticed they are very inconsiatnat in returning the value.

vijayrap:
Added my points, I need ur review on few points, My results are very inconsistent  on day to day basis. I don't know the reason,  Specially, when the CRD updates  the status field , even though the work flow captures "active" it reports and updates as "notActive", on another service, it captures correctly.

test/env/deploy/cluster.yaml
vijayrap: updated

pkg/controller/contrailmonitor/contrailmonitor_controller.go
kaweue: If your code works inconsistently you should write unit tests and e2e tests and fix it. That is test driven development.

PR #304, opened 2020-08-06
Upload crds via init container

kaweue:

PR #311, opened 2020-08-11
Fetch crdsloader image to local registry

michalskalski:


Dysproz:
Hi @michalskalski,
What's the purpose of this image? Why should we download it into local registry?

michalskalski:
Hi @Dysproz it looks like the crdsloader became init container for contrail-operator since this change https://github.com/Juniper/contrail-operator/pull/304 . I tried today setup test env with master branch and my registry did not contain this image.

PR #315, opened 2020-08-12
Change analytics_statistics_ttl to 4

Dysproz:
Change analytics_statistics_ttl to 4 as requested by Yuvaraja


Dysproz:
Waiting for Yuvaraja's confirmation before merge

PR #321, opened 2020-08-18
adding statusmonitor/main.go

vijayrap:
Raising a review request.

### DIFF
--- a/statusmonitor/main.go
+++ b/statusmonitor/main.go
@@ -93,11 +93,13 @@ func main() {
 				}
 				client, err := CreateRestClient(config)
 				if err != nil {
-					log.Fatalf("rest client creation failed: %v", err)
+					log.Printf("rest client creation failed: %v", err)
+                                        continue
 				}
 				clientset, restClient, err := kubeClient(config)
 				if err != nil {
-					log.Fatalf("kubernates client creation failed: %v", err)
+					log.Printf("kubernates client creation failed: %v", err)
+                                        continue
 				}
 				ticker = time.NewTicker(time.Duration(config.Interval) * time.Second)
 				switch config.NodeType {
--- a/statusmonitor/main.go
+++ b/statusmonitor/main.go
@@ -94,17 +94,21 @@ func main() {
 				client, err := CreateRestClient(config)
 				if err != nil {
 					log.Printf("rest client creation failed: %v", err)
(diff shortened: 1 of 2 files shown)


psykulsk (changes requested):
Thanks @vijayrap, that's good, but more changes are required:

This function has to return error
https://github.com/Juniper/contrail-operator/blob/new_2008R_update_CEM-18251/statusmonitor/config_status_monitor.go#L56
Instead of doing log.Fatalf() here https://github.com/Juniper/contrail-operator/blob/new_2008R_update_CEM-18251/statusmonitor/config_status_monitor.go#L89

Same with the getControlStatus, it should return error:
https://github.com/Juniper/contrail-operator/blob/new_2008R_update_CEM-18251/statusmonitor/main.go#L118
instead of log.Fatalf: https://github.com/Juniper/contrail-operator/blob/new_2008R_update_CEM-18251/statusmonitor/main.go#L136
and here https://github.com/Juniper/contrail-operator/blob/new_2008R_update_CEM-18251/statusmonitor/main.go#L123

Then, in the main for loop, you can check those errors and log them here https://github.com/Juniper/contrail-operator/blob/new_2008R_update_CEM-18251/statusmonitor/main.go#L107 and here https://github.com/Juniper/contrail-operator/blob/new_2008R_update_CEM-18251/statusmonitor/main.go#L109 .
Thanks to that, they will be logged but the process will not exit.

statusmonitor/main.go
```diff
@@ -93,16 +93,22 @@ func main() {
...
-					log.Fatalf("kubernates client creation failed: %v", err)
+					log.Printf("kubernates client creation failed: %v", err)
+					continue
 				}
 				ticker = time.NewTicker(time.Duration(config.Interval) * time.Second)
 				switch config.NodeType {
 				case "control":
-					getControlStatus(client, clientset, restClient, config)
+					err := getControlStatus(client, clientset, restClient, config)
+					if err != nil{
```
Dysproz, line 108: It may be done as one liner:
```
if err := getControlStatus(client, clientset, restClient, config); err != nil {
```

Dysproz (approved):
LGTM. One NIT comment

psykulsk (approved):

psykulsk:
recheck

PR #323, opened 2020-08-20
Extend keystone CRD with ExternalAddress option.

michalskalski:
When ExternalAddress is defined for keystone CRD no other actions will be taken by keystone controller other than updating status of the resource with information that it is external resource. This information can be later used by keystone client to decide how to connect to server, either using proxy or direct client. Because new direct connection may require additional steps like including CA to the request, those steps were moved inside keystone client to reduce code needed to initialize new client.


pkg/client/keystone/keystone.go
```diff
@@ -2,20 +2,94 @@ package keystone
...
+		KeystoneConf: &k.Spec.ServiceConfiguration,
+	}, nil
+}
+
+type keystoneClient interface {
+	NewRequest(method, path string, body io.Reader) (*http.Request, error)
+	Do(req *http.Request) (*http.Response, error)
 }
 
 type Client struct {
```
katrybacka, line 52: this is confusing that the field in Client struct has the same name like the struct, I see that it was done like this before but could you please change it to sth like Proxy?

pkg/client/keystone/keystone.go
```diff
@@ -2,20 +2,94 @@ package keystone
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func NewClient(client *kubeproxy.Client) *Client {
-	return &Client{client: client}
+// NewClient prepares keystone client based on properties of passed keystone CRD.
+func NewClient(kubClient client.Client, scheme *runtime.Scheme, config *rest.Config, k *contrail.Keystone) (*Client, error) {
+	if k.Spec.ServiceConfiguration.ExternalAddress != "" {
```
katrybacka, line 23: you could split this method to create Client either external, secure, not-secure then you could return a struct I mean sth like that:
```
func NewClient(kubClient client.Client, scheme *runtime.Scheme, config *rest.Config, k *contrail.Keystone) (*Client, error) {
	proxy, err := newProxy(kubClient, scheme, config, k)
	if err != nil {
		return nil, err
	}
	return &Client{
		Client:       proxy,
		KeystoneConf: &k.Spec.ServiceConfiguration,
	}, nil
}

func newProxy(kubClient client.Client, scheme *runtime.Scheme, config *rest.Config, k *contrail.Keystone) (keystoneClient, error) {
	if k.Spec.ServiceConfiguration.ExternalAddress != "" {
		caCertificate := certificates.NewCACertificate(kubClient, scheme, k, k.GetName())
		caBundle, _ := caCertificate.GetCaCert()
		return newExtKeystoneClient(k.Spec.ServiceConfiguration.AuthProtocol, k.Spec.ServiceConfiguration.ExternalAddress, k.Spec.ServiceConfiguration.ListenPort, caBundle), nil
	}
	proxy, err := kubeproxy.New(config)
	if err != nil {
		return nil, fmt.Errorf("failed to create kubeproxy: %v", err)
	}
	if k.Spec.ServiceConfiguration.AuthProtocol == "https" {
		return proxy.NewSecureClientForService(k.Namespace, k.Name+"-service", k.Status.Port), nil
	}
	return proxy.NewClientForService(k.Namespace, k.Name+"-service", k.Status.Port), nil
}```

Comment (katrybacka) [reply pkg/client/keystone/keystone.go:23]: ofc there might be some errors and you can change names, but my suggestion is to split it to make it more readable

File: pkg/client/keystone/keystone.go [hunk 3]
```diff
@@ -2,20 +2,94 @@ package keystone
...
+		caCertificate := certificates.NewCACertificate(kubClient, scheme, k, k.GetName())
+		caBundle, _ := caCertificate.GetCaCert()
```
katrybacka, line 25: do we want to ignore this error? i think that in case of error we should perform another action like retry since you already created a CACert and it cannot be retrieved or maybe error should be returned

pkg/client/keystone/keystone.go
katrybacka: Moreover why these fields are public?

pkg/controller/keystone/keystone_controller.go
```diff
@@ -296,6 +310,30 @@ func (r *ReconcileKeystone) updateStatus(
...
+}
+
+func (r *ReconcileKeystone) updateStatusWithExternalKeystone(
+	k *contrail.Keystone,
+) error {
+	k.Status = contrail.KeystoneStatus{}
+	k.Status.Active = true
+	k.Status.External = true
+	k.Status.Port = k.Spec.ServiceConfiguration.ListenPort
+	k.Status.ClusterIP = k.Spec.ServiceConfiguration.ExternalAddress
```
katrybacka, line 333: maybe if right now we will support external keystone this field in status should have different name than ClusterIP since it can be some externalIP or even domain name if I understood correctly, so please rename it to Endpoint or sth.

pkg/controller/keystone/keystone_controller.go
```diff
@@ -296,6 +310,30 @@ func (r *ReconcileKeystone) updateStatus(
 	return r.client.Status().Update(context.Background(), k)
 }
 
+func (r *ReconcileKeystone) externalKeystoneReady(k *contrail.Keystone, adminPasswordSecret string) (bool, error) {
+	keystoneClient, err := keystone.NewClient(r.client, r.scheme, r.restConfig, k)
+	if err != nil {
+		return false, err
+	}
+	_, err = keystoneClient.PostAuthTokens("admin", adminPasswordSecret, "admin")
```
katrybacka, line 318: maybe we can introduce some retry mechanism here?

test/e2e/aio/cluster_test.go
```diff
@@ -25,6 +26,7 @@ import (
 	"github.com/Juniper/contrail-operator/pkg/controller/utils"
 	"github.com/Juniper/contrail-operator/test/logger"
 	wait "github.com/Juniper/contrail-operator/test/wait"
+	k8client "sigs.k8s.io/controller-runtime/pkg/client"
```
katrybacka, line 29: sort imports

test/e2e/aio/cluster_test.go
```diff
@@ -113,6 +115,14 @@ func TestCluster(t *testing.T) {
 
 			configProxy := proxy.NewSecureClient("contrail", "config1-config-statefulset-0", 8082)
 
+			keystoneCRD := &contrail.Keystone{}
+			err := f.Client.Get(context.TODO(),
+				types.NamespacedName{
+					Namespace: namespace,
+					Name:      "keystone",
+				}, keystoneCRD)
+			assert.NoError(t, err)
```
katrybacka, line 124: I guess you should require noerror here since without keystone CR it won't work

test/e2e/aio/cluster_test.go
```diff
@@ -113,6 +115,14 @@ func TestCluster(t *testing.T) {
 
 			configProxy := proxy.NewSecureClient("contrail", "config1-config-statefulset-0", 8082)
 
+			keystoneCRD := &contrail.Keystone{}
```
katrybacka, line 118: its not a CRD but CR

test/e2e/aio/command_test.go
```diff
@@ -257,8 +259,19 @@ func TestCommandServices(t *testing.T) {
 				assert.NotEmpty(t, commandPods.Items)
 			})
 
+			keystoneCRD := &contrail.Keystone{}
```
katrybacka, line 262: same here

test/e2e/aio/command_test.go
```diff
@@ -272,8 +285,11 @@ func TestCommandServices(t *testing.T) {
 				assert.NoError(t, err)
 			})
 
-			keystoneProxy := proxy.NewSecureClientForService("contrail", "commandtest-keystone-service", 5555)
-			keystoneClient = keystone.NewClient(keystoneProxy)
+			runtimeClient, err := k8client.New(f.KubeConfig, k8client.Options{Scheme: f.Scheme})
+			assert.NoError(t, err)
```
katrybacka, line 289: require

test/e2e/aio/command_test.go
```diff
@@ -272,8 +285,11 @@ func TestCommandServices(t *testing.T) {
...
+			keystoneClient, err = keystone.NewClient(runtimeClient, f.Scheme, f.KubeConfig, keystoneCRD)
+			assert.NoError(t, err)
```
katrybacka, line 291: require

test/e2e/aio/cluster_test.go
```diff
@@ -122,8 +132,10 @@ func TestCluster(t *testing.T) {
 			})
 
 			t.Run("then config nodes are created", func(t *testing.T) {
-				keystoneProxy := proxy.NewSecureClient("contrail", "keystone-keystone-statefulset-0", 5555)
-				keystoneClient := keystone.NewClient(keystoneProxy)
+				// Test framework client has incompatible Create method signature
+				runtimeClient, err := k8client.New(f.KubeConfig, k8client.Options{Scheme: f.Scheme})
```
katrybacka, line 136: require.NoError is missing

test/e2e/aio/openstack_test.go
```diff
@@ -300,3 +300,24 @@ func TestOpenstackServices(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func getKeystoneClient(f *test.Framework, namespace string, instanceName string) (*keystone.Client, error) {
+	keystoneCRD := &contrail.Keystone{}
```
katrybacka, line 305: CR

pkg/client/keystone/keystone.go
michalskalski: Renamed to Connector if that is ok? Wanted avoid confusion with kubeproxy client.

I made fields public because there is one case in command tests https://github.com/Juniper/contrail-operator/blob/73368d027fe6c60edd2a58772d066943a05c7f8e/test/e2e/aio/command_test.go#L270-L279 where connection to keystone through command pod is verifed. To handle such situation in newClient method of keyston client I would need to define additional parameters. I did not saw usage of it outside of test (rest of the requests goes through keystone service), but wanted to keep that test defined.

test/e2e/aio/cluster_test.go
michalskalski: Added

pkg/client/keystone/keystone.go
michalskalski: Thank you, followed your suggestion.

pkg/client/keystone/keystone.go
michalskalski: My understanding was if the caBundle will be nill (in case of  GetCaCert() error) the newExtKeystoneClient will not add anything to rootCA. If the caBundle is required for endpoint verification the client will return error. If endpoint cert is signed by authority defined in operating system ca bundle then request still should succeed even getting internal ca failed.

pkg/controller/keystone/keystone_controller.go
michalskalski: I don't have too much experience yet with operator, my current understanding is that in case of error reconcile will be repeated, which I saw as some kind of retry mechanism.

pkg/client/keystone/keystone.go
```diff
@@ -2,20 +2,94 @@ package keystone
...
 	"bytes"
+	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 
+	contrail "github.com/Juniper/contrail-operator/pkg/apis/contrail/v1alpha1"
```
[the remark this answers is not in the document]
michalskalski: Grouped imports as you described.

pkg/controller/keystone/keystone_controller.go
michalskalski: Renamed to Endpoint, also for internal variables changed IP references to more general 'address'.

pkg/controller/keystone/keystone_controller.go
katrybacka: You are right I was thinking that connection to some external endpoint can be more flaky than within the cluster so maybe we could check the POST request with a retry mechanism but ofc it's not a bug to return with error from Reconcile and let it be invoked later. That was an idea that came to my mind but it is not a necessary feature, it can stay like that.

test/e2e/aio/command_test.go
michalskalski: Changed

test/e2e/aio/command_test.go
michalskalski: Changed

test/e2e/aio/command_test.go
michalskalski: Changed

test/e2e/aio/cluster_test.go
michalskalski: Sorted

test/e2e/aio/cluster_test.go
michalskalski: changed to require

test/e2e/aio/cluster_test.go
michalskalski: renamed to CR

test/e2e/aio/openstack_test.go
michalskalski: renamed to CR

michalskalski:
recheck

katrybacka:
recheck ha

kaweue:
recheck

pkg/client/keystone/keystone.go
katrybacka: that's completely fine

katrybacka (approved):

PR #324, opened 2020-08-21
Small change in  statusmonitor/main.go

vijayrap:
Small change in  statusmonitor/main.go


vijayrap:
Please check small changes

PR #330, opened 2020-08-25
Add hostname to /etc/hosts on cassandra

mfigurski:
On bare kubernetes cluster cassandra's containers fail with error: `hostname: Name or service not known error`. The reason of that is missing entry in /etc/hosts about hostname stored in /etc/hostname.

### DIFF
--- a/pkg/controller/cassandra/cassandra_controller.go
+++ b/pkg/controller/cassandra/cassandra_controller.go
@@ -167,7 +167,7 @@ type ReconcileCassandra struct {
 }
 
 var cassandraInit2CommandTemplate = template.Must(template.New("").Parse(
-	"keytool -keystore /etc/keystore/server-truststore.jks -keypass {{ .KeystorePassword }} -storepass {{ .TruststorePassword }} -list -alias CARoot -noprompt;" +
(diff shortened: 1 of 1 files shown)


Dysproz (approved):

pkg/controller/cassandra/cassandra_controller.go
```diff
@@ -167,7 +167,7 @@ type ReconcileCassandra struct {
 }
 
 var cassandraInit2CommandTemplate = template.Must(template.New("").Parse(
-	"keytool -keystore /etc/keystore/server-truststore.jks -keypass {{ .KeystorePassword }} -storepass {{ .TruststorePassword }} -list -alias CARoot -noprompt;" +
+	"echo \"$POD_IP $(cat /etc/hostname)\" >> /etc/hosts && keytool -keystore /etc/keystore/server-truststore.jks -keypass {{ .KeystorePassword }} -storepass {{ .TruststorePassword }} -list -alias CARoot -noprompt;" +
```
kaweue, line 170: What was the hostname in this case?

It looks like a hack for me to fix some environment/server configuration issues. I don't like that.

kaweue (changes requested):

PR #331, opened 2020-08-25
Schedule pods only on selected nodes

mfigurski:


pkg/controller/keystone/keystone_controller.go
```diff
@@ -496,7 +496,7 @@ func newKeystoneService(cr *contrail.Keystone) *core.Service {
 	}
 }
 
-func newBootStrapJob(cr *contrail.Keystone, name types.NamespacedName, kcbName, fernetKeysSecretName, credentialKeysSecretName string, psqlIP string) *batch.Job {
+func newBootStrapJob(cr *contrail.Keystone, name types.NamespacedName, kcbName, fernetKeysSecretName, credentialKeysSecretName string, psqlIP string, nodeSelector map[string]string) *batch.Job {
```
katrybacka, line 499: please remove `nodeSelector` argument since you can take it from `cr.CommonConfiguration.NodeSelector`

test/env/deploy/cluster.yaml
```diff
@@ -264,3 +264,10 @@ spec:
                 image: registry:5000/common-docker-third-party/contrail/centos-binary-swift-account:train-2005
               - name: swiftAccountServer
                 image: registry:5000/common-docker-third-party/contrail/centos-binary-swift-account:train-2005
+        commonConfiguration:
```
katrybacka, line 267: why hostNetwork is needed? Move `commonConfiguration` before `serviceConfiguration` like in other components and remove hostNetwork from here it should be set at Manager level

katrybacka (changes requested):

katrybacka (approved):

kaweue (approved):

PR #344, opened 2020-08-27
Increase reliability of e2e tests

kaweue:


kaweue:
recheck ha

kaweue:
Following changes were applied:

- zookeeper nodes are added one by one instead of 2 at the same time since it might cause some issues
```Never specify more than one joining server in the same initial configuration as participants. Currently, the joining servers don’t know that they are joining an existing ensemble; if multiple joiners are listed as participants they may form an independent quorum creating a split-brain situation such as processing operations independently from your main ensemble.```

- retry some of the request to k8s api server since they may fail from time to time under heavy load
- use only one k8s control-plan node since it will consume less resources and it should eliminate some random fails of k8s api (for example in case of etcd leader election
- use zookeeper 3.5.5 since it is more stable

tomaszmj (approved):

shifali19 (approved):

PR #350, opened 2020-08-28
Adding changes for control status-monitor and test-case's

tmallick:
Made changes for control-status monitor to use introspect port instead of analytics. Status-monitor is displayed correctly. 

Added test-case for control-statusmonitor.


tmallick:
Merged using other branch

PR #364, opened 2020-09-08
Keystone object-storage service name parameterized

michalskalski:
When external keystone is in use it can already has a definition of
object-storage service. With SwiftServiceName property it is now
possible to decide about service name created by operator in keystone.
Default value is set to "swift".

### DIFF
--- a/pkg/apis/contrail/v1alpha1/keystone_types.go
+++ b/pkg/apis/contrail/v1alpha1/keystone_types.go
@@ -69,6 +69,33 @@ type KeystoneList struct {
 	Items           []Keystone `json:"items"`
 }
 
+// SetDefaultValues sets default values for keystone resource parameters
+func (k *Keystone) SetDefaultValues() {
+	// After migration to CRD apiextensions.k8s.io/v1 replace those conditions
+	// with +kubebuilder:default markers on keystone struct fileds
+	if k.Spec.ServiceConfiguration.ListenPort == 0 {
+		k.Spec.ServiceConfiguration.ListenPort = KeystoneAuthPublicPort
+	}
+	if k.Spec.ServiceConfiguration.Region == "" {
+		k.Spec.ServiceConfiguration.Region = KeystoneAuthRegionName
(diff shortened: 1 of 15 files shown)


pkg/apis/contrail/v1alpha1/swift_types.go
```diff
@@ -51,6 +51,13 @@ type SwiftList struct {
 	Items           []Swift `json:"items"`
 }
 
+// SetDefaultValues sets default values for swift resource parameters
+func (s *Swift) SetDefaultValues() {
+	if s.Spec.ServiceConfiguration.SwiftProxyConfiguration.SwiftServiceName == "" {
+		s.Spec.ServiceConfiguration.SwiftProxyConfiguration.SwiftServiceName = SwiftServiceName
+	}
+}
```
Dysproz, line 59: So if you use helper function like this then use the same logic for keystone_types.go :)

pkg/apis/contrail/v1alpha1/keystone_types.go
```diff
@@ -69,6 +69,33 @@ type KeystoneList struct {
...
+	if k.Spec.ServiceConfiguration.UserDomainID == "" {
+		k.Spec.ServiceConfiguration.UserDomainID = KeystoneAuthUserDomainID
+	}
+	if k.Spec.ServiceConfiguration.ProjectDomainName == "" {
+		k.Spec.ServiceConfiguration.ProjectDomainName = KeystoneAuthProjectDomainName
+	}
+	if k.Spec.ServiceConfiguration.ProjectDomainID == "" {
+		k.Spec.ServiceConfiguration.ProjectDomainID = KeystoneAuthProjectDomainID
+	}
+}
```
Dysproz, line 97: Use similar helper function like in swift_types :)

Dysproz (changes requested):

Dysproz (approved):

kaweue (approved):

PR #379, opened 2020-09-16
 added files to master with updated structures in most of the package…

vijayrap:
added files to master with updated structures in most of the packages like Podconfiguration etc

All the changes in the related to R2008 are their, except that in the test cases there is some major updates in master as per the HA requirements.

Request to check


rkoverzin (approved):

lrebeja (approved):
As discussed approving without UTs and expect it next review.

psykulsk:
@vijayrap please use the squash and merge option when merging pull requests. It makes the commit history more clean. Now  we have multiple commits in the master history that say things like: "fix", "add file", "fix" etc.

PR #382, opened 2020-09-17
Vrouter improvements

psykulsk:
This PR modifies the spec of vrouter pod in a way that the PHYSICAL_INTERFACE environment variable is set using a reference to annotation instead of taking a value from a ConfigMap. This way, each pod can have a different value of this env var. Previously all pods under the same vrouter custom resource had to use the same value of this env var, so all nodes had to have the same name of the physical interface.

Additionally there is the option to set the physicalInterface via service configuration. It then overrides the value set by the annotation with a value from the config map. I'm not sure about the use-case of it (maybe when nodes have multiple interfaces), but it was there before, so I left it.

I also added a pod anti-affinity so that two vrouter pods will not be scheduled on the same node. This could happen for example if a node would have both the master and worker label and there were to vrouter CRs, one with nodeSelector for master label, and another with nodeSelector for the worker label.

### DIFF
--- a/pkg/apis/contrail/v1alpha1/vrouter_types.go
+++ b/pkg/apis/contrail/v1alpha1/vrouter_types.go
@@ -325,7 +325,6 @@ func (c *Vrouter) InstanceConfiguration(request reconcile.Request,
 	sort.SliceStable(podList.Items, func(i, j int) bool { return podList.Items[i].Status.PodIP < podList.Items[j].Status.PodIP })
 	var data = make(map[string]string)
 	for idx := range podList.Items {
-		// GENERATE configmaps peor pod (at least the one with ENV vars)
 		hostname := podList.Items[idx].Annotations["hostname"]
 		physicalInterfaceMac := podList.Items[idx].Annotations["physicalInterfaceMac"]
 		prefixLength := podList.Items[idx].Annotations["prefixLength"]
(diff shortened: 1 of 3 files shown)


pkg/apis/contrail/v1alpha1/vrouter_types.go
```diff
@@ -134,6 +134,19 @@ func (c *Vrouter) PrepareDaemonSet(ds *appsv1.DaemonSet,
...
+				LabelSelector: &metav1.LabelSelector{
+					MatchExpressions: []metav1.LabelSelectorRequirement{{
+						Key:      instanceType,
+						Operator: "Exists",
+					}},
+				},
+				TopologyKey: "kubernetes.io/hostname",
+			}},
+		},
+	}
```
Dysproz, line 149: Does it make sense since DaemonSet runs single replica on every node?

Dysproz (changes requested):

pkg/apis/contrail/v1alpha1/vrouter_types.go
psykulsk: Yes because on Openshift we have 2 vrouter daemonsets. Have a look at the explanation in description.

Dysproz (approved):

mfigurski (approved):

PR #394, opened 2020-09-22
Refactor config generation

psykulsk:
I modified the ControlClusterConfiguration and ConfigClusterConfiguration structures. Before, different strings containing formatted lists of IPs or endpoints were stored in the struct (comma separated, space separated, comma separated with quotes etc.).  

I modified it so that only the actual data is stored in the struct and strings in the required formats are generated just before the template is executed (using helper functions), so only the relevant data is passed in the struct. 

I believe that it simplifies the code and will also make it easier to add the static configuration capabilities to different custom resources.

In future PRs other ClusterConfiguration structs will be similarly refactored.

### DIFF
--- a/pkg/apis/contrail/v1alpha1/templates/templates_helper_functions.go
+++ b/pkg/apis/contrail/v1alpha1/templates/templates_helper_functions.go
@@ -28,6 +28,9 @@ func EndpointList(ips []string, port int) []string {
 }
 
 func EndpointListCommaSeparated(ips []string, port int) string {
+	if len(ips) == 0 {
+		return ""
+	}
 	portStr := strconv.Itoa(port)
 	endpointList := strings.Join(ips, ":"+portStr+",")
 	endpointList = endpointList + ":" + portStr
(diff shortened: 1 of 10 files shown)


pkg/apis/contrail/v1alpha1/templates/templates_helper_functions.go
```diff
@@ -0,0 +1,49 @@
+package templates
```
Dysproz, line 1: Cannot these separate functions be more universal?
For example I've created something like that:
```
func joinServerList(servers []string, port *int, sep string) string {
	if port != nil {
		strPort := strconv.Itoa(*port)
		return strings.Join(servers, ":"+strPort+sep) + ":" + strPort
	}
	return strings.Join(servers, sep)

```

Because in case that we'd need to use different separator or sth then we'll have to make new function

psykulsk: Done. Left the separate functions, but they both call a common function that accepts the separator as the argument.

shifali19 (approved):

psykulsk:
recheck

Dysproz:
Still there are separate methods for everything :<

Dysproz (approved):

PR #399, opened 2020-09-25
Fix command configuration file

shifali19:


kabart (approved):

pkg/k8s/config_map.go
```diff
@@ -33,7 +33,9 @@ func (c *ConfigMap) EnsureExists(dataSetter configMapFiller) error {
 		return err
 	}
 	_, err = controllerutil.CreateOrUpdate(context.Background(), c.client, cm, func() error {
-		cm.Data = map[string]string{}
+		if len(cm.Data) == 0 {
```
katrybacka, line 36: I think the change should be not here but in generation of the config map in Command since right now  i see that we are invoking EnsureExists method on config map which actually appends some data to the config map instead of non-incrementaly change it. The question is how this can affect other config maps? and moreover when I see that we are ensuring that a specific config map exist I understand it like the content of it will be how I see it in dataSetter. I'm not saying its a bad solution because it is working but I would suggest  not to change the interface of our k8s.ConfigMap but instead in command controller iterate over pods fill the Config map correctly for all of them and update the config map once. It will be even faster since we will reduce number of API calls.

katrybacka (changes requested):

shifali19:
recheck

pkg/k8s/config_map.go
shifali19: Done

pkg/controller/command/command_config.go
```diff
@@ -23,11 +37,30 @@ type commandConf struct {
...
+			CAFilePath:      c.CAFilePath,
+			PGPassword:      c.PGPassword,
+		}
+		conf.fillConfigMapForPod(cm)
+	}
+}
+
+func (c *commandPodConf) fillConfigMapForPod(cm *core.ConfigMap) {
 	cm.Data["command-app-server"+c.HostIP+".yml"] = c.executeTemplate(commandConfig)
 	cm.Data["entrypoint.sh"] = c.executeTemplate(commandEntrypoint)
```
katrybacka, line 60: entrypoint.sh is parameterized too and I see that it is overwritten here

katrybacka (changes requested):

pkg/controller/command/command_config.go
katrybacka: Now it's good

katrybacka (approved):

shifali19:
recheck

shifali19:
recheck

shifali19:
recheck

PR #402, opened 2020-09-28
Fix contrailCNI active status

Dysproz:
Quick fix for ContrailCNI active status which was not set to true after job complete.
Also removed unnecessary label for ContrailCNI from manager.


pkg/controller/manager/manager_controller.go
```diff
@@ -674,9 +674,6 @@ func (r *ReconcileManager) processContrailCNIs(manager *v1alpha1.Manager) error
 			oldContrailCNI.ObjectMeta = v1.ObjectMeta{
 				Namespace: manager.Namespace,
 				Name:      *existingContrailCNI.Name,
-				Labels: map[string]string{
-					"contrail_cluster": "cluster1",
-				},
```
psykulsk, line 679: Now it will delete any ContrailCNI object in the namespace. Shouldn't it delete only the one in the same "contrail_cluster" as the current Manager?

psykulsk (changes requested):

pkg/controller/manager/manager_controller.go
Dysproz: Done - for all resources

psykulsk (approved):

pkg/controller/manager/manager_controller.go
```diff
@@ -381,6 +387,9 @@ func (r *ReconcileManager) processWebui(manager *v1alpha1.Manager, replicas *int
 			oldWebUI.ObjectMeta = v1.ObjectMeta{
 				Namespace: manager.Namespace,
 				Name:      *manager.Status.Webui.Name,
+				Labels: map[string]string{
```
kaweue, line 390: Why is this required?

Dysproz: because manager adds label with its name to created resources which points to this manager as owner. So in case there would be other WebUI created by e.g. *cluster2* manager and we process this Delete without labels from *cluster1* manager then it'll delete *cluster2's* WebUI. Labels will protect resources created by other managers

kaweue: smart

kaweue (approved):

PR #405, opened 2020-09-29
Remove invalid owner references

psykulsk:
Resources created by our controllers are still sometimes being deleted by the kubernetes garbage collector. The cause is the same as with this PR https://github.com/Juniper/contrail-operator/pull/277 , please refer to the description there. But this time the owner references are set on the ClusterRole and ClusterRoleBinding objects which are cluster-scoped, so they can't have an namespace scoped owner like our CustomResources.

To trigger the garbage collection kill the pid 1 process inside all kube-controller-manager containers:
On Openshift:
```
kubectl get pod -n openshift-kube-controller-manager -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' | grep kube-controller-manager-ip | xargs -I% kubectl exec -n openshift-kube-controller-manager % bash -- -c 'kill 1'
```

On kind (do it for all replicas):
```
kubectl exec -it -n kube-system     kube-controller-manager-kind-control-plane sh -- -c 'kill 1'
```

### DIFF
--- a/pkg/apis/contrail/v1alpha1/base_types.go
+++ b/pkg/apis/contrail/v1alpha1/base_types.go
@@ -206,10 +206,6 @@ func CreateAccount(accountName string, namespace string, client client.Client, s
 				},
 			}},
 		}
-		err = controllerutil.SetControllerReference(owner, clusterRole, scheme)
-		if err != nil {
-			return err
-		}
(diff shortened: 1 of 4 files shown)


psykulsk:
recheck ha

Dysproz (approved):

kaweue (approved):

PR #409, opened 2020-10-07
Add hostnames of other nodes to host aliases

kaweue:


kaweue:
recheck

Dysproz:
recheck ha

deploy/1-create-operator.yaml
```diff
@@ -79,7 +79,7 @@ rules:
 - apiGroups:
   - contrail
   resources:
-  - '*'
+  - contrailcnis
```
Dysproz, line 82: Why this change?

pkg/apis/contrail/v1alpha1/base_types.go
```diff
@@ -100,6 +100,12 @@ type PodConfiguration struct {
 	// +k8s:conversion-gen=false
 	// +optional
 	HostNetwork *bool `json:"hostNetwork,omitempty" protobuf:"varint,11,opt,name=hostNetwork"`
+	// HostAliases is an optional list of hosts and IPs that will be injected into the pod's hosts
+	// file if specified. This is only valid for non-hostNetwork pods.
```
katrybacka, line 104: non-hostNetwork pods?

kaweue: It looks like this comment is not true. It was copy passed from core/v1/types.go.
However this is supported for hostnetworking:
https://github.com/kubernetes/kubernetes/commit/4edd92f26d812c2bc7b7442e6a553ca3f60a4a8e

kaweue: I will delete that sentence from this comment

deploy/1-create-operator.yaml
kaweue: I just run ./create_manifest.sh

kaweue: I will change that in deploy/role.yaml

kaweue:
recheck ha

pkg/apis/contrail/v1alpha1/base_types.go
kaweue: done

deploy/1-create-operator.yaml
kaweue: done

katrybacka (approved):

psykulsk:
recheck ha

Dysproz (approved):

PR #415, opened 2020-10-09
Master node failover fix, cassandra replication_factor

psykulsk:
Replication factor setting in cassandra is based on the number of nodes specified in the config api configuration files.  When a single ClusterIP was used, the replication_factor for the contrail keyspaces in cassandra was set to 1. This caused kubemanager and control failures in an HA (3 master nodes) setup when one of the master nodes was turned off.

We have to go back to using the list of cassandra IPs/endpoints until we figure out another way.


kaweue (approved):

kaweue:
Maybe it will fix this test case in e2e tests
 https://github.com/Juniper/contrail-operator/blob/e05ff1ab624590a89d1bd8b624edbbf2496ffde6/test/e2e/ha/core_contrail_services_test.go#L190?

katrybacka (approved):

Dysproz (approved):
I'm not convinced and would like it to be solved better in contrail, but well... LGTM! ;)

psykulsk:
recheck ha

kaweue (approved):

PR #417, opened 2020-10-12
Migrate kubemanager and vrouter dynamic configuration to manager

Dysproz:


Dysproz:
recheck

Dysproz:
recheck

pkg/apis/contrail/v1alpha1/kubemanager_types.go
```diff
@@ -113,17 +107,21 @@ func (c *Kubemanager) InstanceConfiguration(request reconcile.Request,
...
-	cassandraNodesInformation, err := c.getCassandraNodesInformation(request.Namespace, client)
-	configNodesInformation, err := c.getConfigNodesInformation(request.Namespace, client)
-	rabbitmqNodesInformation, err := c.getRabbitmqNodesInformation(request.Namespace, client)
-	zookeeperNodesInformation, err := c.getZookeeperNodesInformation(request.Namespace, client)
+	cassandraNodesInformation := c.Spec.ServiceConfiguration.CassandraNodesConfiguration
+	cassandraNodesInformation.FillWithDefaultValues()
+	configNodesInformation := c.Spec.ServiceConfiguration.ConfigNodesConfiguration
+	configNodesInformation.FillWithDefaultValues()
+	rabbitmqNodesInformation := c.Spec.ServiceConfiguration.RabbbitmqNodesConfiguration
+	rabbitmqNodesInformation.FillWithDefaultValues()
```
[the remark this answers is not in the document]
Dysproz: Yes, because user may provide selected configuration in static mode and the rest is expected ot be filled with default values.

pkg/apis/contrail/v1alpha1/base_types.go
```diff
@@ -101,7 +101,7 @@ type PodConfiguration struct {
 	// +optional
 	HostNetwork *bool `json:"hostNetwork,omitempty" protobuf:"varint,11,opt,name=hostNetwork"`
 	// HostAliases is an optional list of hosts and IPs that will be injected into the pod's hosts
-	// file if specified.
+	// file if specified. This is only valid for non-hostNetwork pods.
```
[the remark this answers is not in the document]
Dysproz: Ok, I removed it, because I wasn't sure whether it should stay or not (all crds had this in description so apparently someone didn't generate crds before commit)

pkg/apis/contrail/v1alpha1/vrouter_types.go
```diff
@@ -56,26 +56,19 @@ type VrouterSpec struct {
...
-	ContrailStatusImage string                      `json:"contrailStatusImage,omitempty"`
-	StaticConfiguration *VrouterStaticConfiguration `json:"staticConfiguration,omitempty"`
-}
-
-// VrouterStaticConfiguration stores static configuration of services
-// required by Vrouter
-// +k8s:openapi-gen=true
-type VrouterStaticConfiguration struct {
+	Containers                []*Container                 `json:"containers,omitempty"`
+	ControlInstance           string                       `json:"controlInstance,omitempty"`
```
kaweue, line 60: Are those fields "ControlInstanc" and "Cassandra" going to be used in vrouter controller?

deploy/crds/contrail.juniper.net_managers_crd.yaml
```diff
@@ -1496,12 +1496,60 @@ spec:
                             properties:
                               cassandraInstance:
                                 type: string
+                              cassandraNodesConfiguration:
```
kaweue, line 1493: In my opinion static configuration shouldn't be a part of manager crds.

pkg/apis/contrail/v1alpha1/vrouter_types.go
Dysproz: In vRouter controller it's not used. I was thinking on dropping them, but in Manager we may declare multiple cassandras and multiple controls resources so we cannot automatically assign proper cassandra and control to vrouter as we cannot say which vrouter should use which cassandra.

deploy/crds/contrail.juniper.net_managers_crd.yaml
Dysproz: Should it be put aside of CommonConfiguration and ServiceConfiguration as separate thing in vRouter spec?

pkg/apis/contrail/v1alpha1/vrouter_types.go
kaweue: I think that this field should stay in manager, but shouldn't be propagated to vrouter. Maybe it is time for a new version of API.
In other words. Manager struct should stay the same, however we may change vrouter type

kaweue: The same is for kubemanager.

kaweue: Moreover CassandraInstance in Vrouter looks totally unused in Vrouter and Manager

deploy/crds/contrail.juniper.net_managers_crd.yaml
kaweue: Yes

kaweue: This is how I see it

kaweue: Actually this ServiceConfiguration is the same struct in Manager and vRouter, it doesn't need to be like this.

Dysproz:
recheck

deploy/crds/contrail.juniper.net_managers_crd.yaml
Dysproz: Ok, I moved static configuration out of manager and as separate struct field in spec

pkg/apis/contrail/v1alpha1/vrouter_types.go
Dysproz: Ok, I redesigned the API

Dysproz:
recheck

Dysproz:
recheck

Dysproz:
recheck

Dysproz:
recheck

pkg/apis/contrail/v1alpha1/kubemanager_types.go
```diff
@@ -54,32 +55,29 @@ type KubemanagerStatus struct {
...
+	IPFabricSnat          *bool        `json:"ipFabricSnat,omitempty"`
+	KubernetesTokenFile   string       `json:"kubernetesTokenFile,omitempty"`
+	HostNetworkService    *bool        `json:"hostNetworkService,omitempty"`
+	RabbitmqUser          string       `json:"rabbitmqUser,omitempty"`
+	RabbitmqPassword      string       `json:"rabbitmqPassword,omitempty"`
+	RabbitmqVhost         string       `json:"rabbitmqVhost,omitempty"`
 }
 
-// KubemanagerStaticConfiguration is the configuration for deployment with static controller.
+// KubemanagerConfiguration is the Spec for the kubemanagers API.
```
kaweue, line 80: This comment is about KubemanagerConfiguration however it should be for KubemanagerStaticConfiguration

pkg/apis/contrail/v1alpha1/kubemanager_types.go
```diff
@@ -37,8 +37,9 @@ type Kubemanager struct {
 // KubemanagerSpec is the Spec for the kubemanagers API.
 // +k8s:openapi-gen=true
 type KubemanagerSpec struct {
-	CommonConfiguration  PodConfiguration         `json:"commonConfiguration,omitempty"`
-	ServiceConfiguration KubemanagerConfiguration `json:"serviceConfiguration"`
+	CommonConfiguration  PodConfiguration               `json:"commonConfiguration,omitempty"`
+	ServiceConfiguration KubemanagerConfiguration       `json:"serviceConfiguration"`
+	StaticConfiguration  KubemanagerStaticConfiguration `json:"staticConfiguration"`
```
kaweue, line 42: In my opinion we should have here just one field `serviceConfiguration`. I don't se the reason why it is split into two fields from the api perspective. We still can have 2 separate structures however it should not be visible from the client perspective.

Moreover I don't like this name `KubemanagerStaticConfiguration` it says nothing, but we know it is just for store information about required services.

pkg/apis/contrail/v1alpha1/manager_types.go
```diff
@@ -23,22 +23,53 @@ type ManagerSpec struct {
...
+	CommonConfiguration  PodConfiguration     `json:"commonConfiguration,omitempty"`
+	ServiceConfiguration VrouterConfiguration `json:"serviceConfiguration"`
+}
+
+// KubemanagerService defines desired confgiuration of vRouter
+// +k8s:openapi-gen=true
+type KubemanagerService struct {
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              KubemanagerServiceSpec `json:"spec,omitempty"`
+	CassandraInstance string                 `json:"cassandraInstance,omitempty"`
```
kaweue, line 64: It looks like manager api is changed by this code. CassandraInstance and ZookeeperInstance should was in spec.
It looks like a breaking change.
We should not introduce breaking changes to manager struct.

kaweue (changes requested):

pkg/apis/contrail/v1alpha1/kubemanager_types.go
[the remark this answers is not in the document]
katrybacka: ok

pkg/apis/contrail/v1alpha1/manager_types.go
```diff
@@ -23,22 +23,53 @@ type ManagerSpec struct {
...
+type VrouterServiceSpec struct {
+	CommonConfiguration  PodConfiguration     `json:"commonConfiguration,omitempty"`
+	ServiceConfiguration VrouterConfiguration `json:"serviceConfiguration"`
+}
+
+// KubemanagerService defines desired confgiuration of vRouter
+// +k8s:openapi-gen=true
+type KubemanagerService struct {
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              KubemanagerServiceSpec `json:"spec,omitempty"`
```
katrybacka, line 63: same concern like Pawel, that we should not break manager, regarding other CR the question is open if we should be able to break or not but for sure you should manipulate somehow these structures not to break the manager structure. Every breaking change should introduce a new version of API

katrybacka (changes requested):

pkg/apis/contrail/v1alpha1/kubemanager_types.go
katrybacka: :+1:

pkg/apis/contrail/v1alpha1/kubemanager_types.go
Dysproz: done

pkg/apis/contrail/v1alpha1/kubemanager_types.go
Dysproz: Ok, So right now Manager's API is the same as before - controlInstance, cassandraInstance etc. are in the same place. But I've hidden static configuration from Manager and left it only in vRouter and Kubemanager API

pkg/apis/contrail/v1alpha1/manager_types.go
Dysproz: it's back in spec

pkg/apis/contrail/v1alpha1/manager_types.go
Dysproz: done

kaweue (approved):

katrybacka (approved):

PR #419, opened 2020-10-13
Command data migration during upgrade

katrybacka:


katrybacka:
recheck

katrybacka:
recheck

kabart (approved):

pkg/controller/command/command_upgrade.go
```diff
@@ -1,26 +1,45 @@
...
-			commandCR.Status.UpgradeState = contrail.CommandStartingUpgradedDeployment
-		} else {
-			currentDeployment.Spec.Template.Spec.Containers = oldDeploymentSpec.Template.Spec.Containers
-			currentDeployment.Spec.Template.Spec.InitContainers = oldDeploymentSpec.Template.Spec.InitContainers
-			currentDeployment.Spec.Replicas = int32ToPtr(0)
+			commandCR.Status.UpgradeState = contrail.CommandUpgrading
+		}
+	case contrail.CommandUpgrading:
+		currentDeployment.Spec.Replicas = int32ToPtr(0)
+		oldImage := getOldCommandImage(oldDeploymentSpec.Template.Spec.Containers, "command")
```
kaweue, line 33: Is it possible to change name of the container is deployment and container is CR?

pkg/controller/command/command_upgrade.go
```diff
@@ -30,8 +49,156 @@ func performUpgradeStepIfNeeded(commandCR *contrail.Command, currentDeployment *
...
+			},
+			TTLSecondsAfterFinished: nil,
+		},
+	}
+}
+
+const dataDumpScript = `
+#!/bin/bash
+
+commandutil convert --intype rdbms --outtype yaml --out /backups/db.yml -c /etc/contrail/command-app-server.yml
```
kaweue, line 192: This script is so small that it could be just an array of strings

pkg/controller/command/command_upgrade.go
```diff
@@ -30,8 +49,156 @@ func performUpgradeStepIfNeeded(commandCR *contrail.Command, currentDeployment *
...
+`
+
+const dataMigrationScript = `
+#!/bin/bash
+
+commandutil migrate --in /backups/db.yml --out /backups/db_migrated.yml
```
kaweue, line 198: What happen when first command fail? Will the next command will be executed?

kaweue: What do you think about having a single command per one container (2 init containers and 1 container)?

pkg/controller/command/command_upgrade.go
```diff
@@ -30,8 +49,156 @@ func performUpgradeStepIfNeeded(commandCR *contrail.Command, currentDeployment *
...
+	}
+	pp := meta.DeletePropagationForeground
+	err := r.client.Delete(context.Background(), job, &client.DeleteOptions{PropagationPolicy: &pp})
+	if !errors.IsNotFound(err) {
+		return err
+	}
+	return nil
 }
 
+func getOldCommandImage(list []core.Container, name string) string {
```
kaweue, line 70: This function only iterates over list of containers and look for the container image. It can be used for any image not only "old command"?

I prefer 'containers' over 'list' as argument

pkg/controller/command/command_controller.go
```diff
@@ -357,8 +374,7 @@ func (r *ReconcileCommand) Reconcile(request reconcile.Request) (reconcile.Resul
 		if err := r.prepareIntendedDeployment(deployment, command); err != nil {
 			return err
 		}
-		performUpgradeStepIfNeeded(command, deployment, oldDeploymentSpec)
-		return nil
+		return r.performUpgradeStepIfNeeded(command, deployment, oldDeploymentSpec, commandBootStrapConfigName)
```
kaweue, line 377: [Not for this PR] I don't really understand why this performUpgradeStepIfNeeded method is inside  this `CreateOrUpdate`. When I look at this I think it only makes things harder. Original deployment needs to be copied, in some states we create a new deployment and then it is overwritten by old configurations...

kaweue (changes requested):

pkg/controller/command/command_upgrade.go
```diff
@@ -30,8 +49,156 @@ func performUpgradeStepIfNeeded(commandCR *contrail.Command, currentDeployment *
...
+							}, {
+								Name:      "backup-volume",
+								MountPath: "/backups/",
+							}},
+						},
+					},
+					DNSPolicy: core.DNSClusterFirst,
+					Tolerations: []core.Toleration{
+						{Operator: "Exists", Effect: "NoSchedule"},
+						{Operator: "Exists", Effect: "NoExecute"},
```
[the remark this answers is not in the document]
katrybacka: right, done

pkg/controller/command/command_upgrade.go
```diff
@@ -30,8 +49,156 @@ func performUpgradeStepIfNeeded(commandCR *contrail.Command, currentDeployment *
...
+							},
+						},
+					},
+					InitContainers: []core.Container{
+						{
+							Name:            "command-upgrade-db-dump",
+							ImagePullPolicy: core.PullAlways,
+							Image:           oldImage,
+							Command:         []string{"bash", "-c", dataDumpScript},
+							VolumeMounts: []core.VolumeMount{{
```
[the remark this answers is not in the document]
katrybacka: done

pkg/controller/command/command_upgrade.go
```diff
@@ -30,8 +49,156 @@ func performUpgradeStepIfNeeded(commandCR *contrail.Command, currentDeployment *
...
+		Spec: batch.JobSpec{
+			BackoffLimit: int32ToPtr(5),
+			Template: core.PodTemplateSpec{
+				Spec: core.PodSpec{
+					HostNetwork:   true,
+					RestartPolicy: core.RestartPolicyNever,
+					NodeSelector:  commandCR.Spec.CommonConfiguration.NodeSelector,
+					Volumes: []core.Volume{
+						{
+							Name: configMapName,
```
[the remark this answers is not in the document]
katrybacka: i'd prefer to propagate it from controller since what about situation when someone will change the configuration name in controller and did not know that it is used here, since this config map was created for another job, here I reuse that

pkg/controller/command/command_upgrade.go
katrybacka: ok I can change that

pkg/controller/command/command_controller.go
katrybacka: unfortunately yes, i'm trying to figure out how to do it better way [not in this PR] :)

pkg/controller/command/command_upgrade.go
katrybacka: changed

pkg/controller/command/command_upgrade.go
katrybacka: done

pkg/controller/command/command_upgrade.go
katrybacka: unified

kaweue (approved):

katrybacka:
recheck

katrybacka:
recheck

katrybacka:
recheck ha

katrybacka:
recheck ha

PR #422, opened 2020-10-14
Enhance command controller

katrybacka:


katrybacka:
recheck

pkg/controller/command/command_controller.go
```diff
@@ -306,33 +285,99 @@ func (r *ReconcileCommand) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 
-	configVolumeName := request.Name + "-" + instanceType + "-volume"
-	csrSignerCaVolumeName := request.Name + "-csr-signer-ca"
+	deployment := newDeploymentWithVolumes(command.Name, command.Namespace, instanceType, command.Spec.ServiceConfiguration.Containers)
+	expectedDeployment := deployment.DeepCopy()
+	performUpgrade := false
+	createOrUpdateResult, err := controllerutil.CreateOrUpdate(context.Background(), r.client, deployment, func() error {
+		if imageIsUpgraded(expectedDeployment, &deployment.Spec) || command.Status.UpgradeState == contrail.CommandStartingUpgradedDeployment {
```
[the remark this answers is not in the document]
kaweue: In my opinion
- prepareIntendedDeployment method should be called only in one place in code
- performUpgradeStep method shouldn't read deployment on its own (it looks like now we read it twice during the one reconciliation)
- performUpgradeStep shouldn't modify the deployment only state of Command CR

pkg/controller/command/command_upgrade.go
```diff
@@ -16,40 +16,59 @@ import (
...
-		expectedReplicas := ptrToInt32(currentDeployment.Spec.Replicas, 1)
-		if currentDeployment.Status.ReadyReplicas == expectedReplicas {
-			commandCR.Status.UpgradeState = contrail.CommandNotUpgrading
+		expectedDeployment.Spec.DeepCopyInto(&d.Spec)
+		d.Spec.Replicas = command.Spec.CommonConfiguration.Replicas
+		if err := r.prepareIntendedDeployment(d, command); err != nil {
+			return nil, err
+		}
+		expectedReplicas := ptrToInt32(d.Spec.Replicas, 1)
+		if d.Status.ReadyReplicas == expectedReplicas {
```
kaweue, line 61: Is it possible that we will have more than one ReadyReplica in that state?

pkg/controller/command/command_upgrade.go
```diff
@@ -76,6 +95,19 @@ func getContainerImage(containers []core.Container, name string) string {
 	return ""
 }
 
+func (r *ReconcileCommand) checkIfImageChangedAgain(command *contrail.Command) (bool, error) {
```
kaweue, line 98: Maybe information about to which version we are upgrading should be in the status of Command.

pkg/controller/command/command_upgrade.go
```diff
@@ -76,6 +95,19 @@ func getContainerImage(containers []core.Container, name string) string {
...
+	}
+	if err != nil {
+		return false, err
+	}
+	return dataMigrationJob.Spec.Template.Spec.Containers[0].Image != getImage(command.Spec.ServiceConfiguration.Containers, "api"), nil
+}
+
 func (r *ReconcileCommand) checkDataMigrationJobStatus(command *contrail.Command) error {
 	dataMigrationJob := &batch.Job{}
 	jobName := types.NamespacedName{Namespace: command.Namespace, Name: command.Name + "-upgrade-job"}
```
kaweue, line 113: Should this "check" method create migrationJob `r.client.Create(context.Background(), dataMigrationJob)`

pkg/controller/command/command_controller.go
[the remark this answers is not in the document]
kaweue: [Optional idea]
Moreover I would also go into completely deferent approach. More object oriented one.
Object CommandCR could have a method/methods that would be responsible for changing the inner state of object. Something like:
- reconcileUpgradeState(deploymentImage string) - that would internally change the state of CR
- upgradeJobComplated()
- getDeploymentSpec()
- getUpgradeState()

kaweue (changes requested):

pkg/controller/command/command_upgrade.go
katrybacka: no it should not, copy paste mistake

pkg/controller/command/command_upgrade.go
katrybacka: yes, we are in that state until whole deployment is ready

pkg/controller/command/command_controller.go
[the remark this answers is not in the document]
katrybacka: i'm not sure about this idea of not modifying deployment in performupgrade state just command cr, i have to think about that. i'll think about splitting logic of modifying cr and deployment

kaweue (approved):

kabart (approved):

PR #424, opened 2020-10-16
Support keystone public ip

kaweue:


pkg/controller/keystone/keystone_config_maps.go
```diff
@@ -32,10 +32,18 @@ func (c *configMaps) ensureKeystoneExists(postgresNode, memcachedNode string, po
 	return c.cm.EnsureExists(cc)
 }
 
-func (c *configMaps) ensureKeystoneInitExist(postgresNode, memcachedNode string, podIP string) error {
+func (c *configMaps) ensureKeystoneInitExist(postgresNode, memcachedNode string, podIP string, nodePort int32) error {
+	publicAddress := podIP
```
katrybacka, line 36: I was so confused by podIP, why we are using pod instead of clusterIP and now i see that the name of this argument is wrong, can we change that to clusterIP?

katrybacka (changes requested):

kabart (approved):

pkg/controller/keystone/keystone_config_maps.go
kaweue: It was like this, but I will change that

katrybacka: yes i see that it is not your change but pls fiiix

katrybacka (approved):

PR #426, opened 2020-10-19
Provision manager read hostname from node not from pod's annotations

kaweue:
Hostname in pod's annotation might not be immediately available, that cause a situation were a dummy (with empty name) analytics or config nodes were added to vnc-api


kabart (approved):

kaweue:
recheck ha

katrybacka (approved):
this method which returns empty map (getPodInitStatus) in case of error confused me, since it get something and in a meantime modifies (wrr). But the PR is gooooood

PR #435, opened 2020-10-27
Rollback after unsuccessful upgrade of Command

kabart:
There should be a logic in Contrail Command Controller that will
rollback the upgrade in case of unsuccessful data migration.


kaweue:
This change fixes migration of Command's data during the upgrade, and also introduces rollback mechanism.
The idea behind this change is that during the migration a new DATABASE is created. This DATABASE is initialised with a new schema and then filled with migrated data. When the process is successful migrated DATABASE is renamed and put in the place of original one. The original one is saved as backup.

pkg/controller/command/command_upgrade.go
```diff
@@ -64,40 +84,45 @@ func (r *ReconcileCommand) deleteMigrationJob(commandCR *contrail.Command) error
...
 func isImageChanged(command *contrail.Command) bool {
 	return command.Status.ContainerImage != "" && command.Status.ContainerImage != getImage(command.Spec.ServiceConfiguration.Containers, "api")
 }
 
-func (r *ReconcileCommand) checkDataMigrationCompleted(command *contrail.Command) (bool, error) {
+func isTargetImageChanged(command *contrail.Command) bool {
+	return command.Status.TargetContainerImage != "" && command.Status.TargetContainerImage != getImage(command.Spec.ServiceConfiguration.Containers, "api")
+}
+
+func (r *ReconcileCommand) checkDataMigrationCompleted(command *contrail.Command) (success bool, failed bool, err error) {
```
katrybacka, line 95: i would prefer to check if the job finished then the second boolean can say if the result was success or not, so the return values would be (finished, succeeded, err) because success and failure are opposites that can be expressed using one variable and for me it will look better if we first ensure that the job is finished then we can check the result.

pkg/apis/contrail/v1alpha1/command_types.go
```diff
@@ -81,6 +83,16 @@ func (c *Command) PodsCertSubjects(podList *corev1.PodList, serviceIP string) []
 	return PodsCertSubjects(podList, c.Spec.CommonConfiguration.HostNetwork, altIPs)
 }
 
+//Upgrading is used to check if command is in the upgrade process
+func (c *Command) Upgrading() bool {
+	if c.Status.UpgradeState == CommandShuttingDownBeforeUpgrade ||
+		c.Status.UpgradeState == CommandUpgrading ||
+		c.Status.UpgradeState == CommandUpgradeFailed {
```
katrybacka, line 90: if upgrade failed then Command is still upgrading? Isn't failure of the upgrade the end of the upgrade?

pkg/controller/command/command_config_bootstrap.go
```diff
@@ -266,3 +268,48 @@ resources:
...
+log_level: debug`))
+
+var migrationScriptConfig = template.Must(template.New("").Parse(`
+#!/usr/bin/env bash
+
+set -e
+export PGPASSWORD={{ .PGPassword }}
+
+# Try to drop old databases the may or may not exists.
+psql -w -h {{ .PostgresAddress }} -U {{ .PostgresUser }} -d {{ .PostgresDBName }} <<END_OF_SQL
```
katrybacka, line 293: [nit] just a nit if you are exporting PGPASSWORD maybe it will be convenient  to store PostgresAddress and PostgresUser in env too: https://www.postgresql.org/docs/9.3/libpq-envars.html

pkg/controller/command/command_config_bootstrap.go
```diff
@@ -266,3 +268,48 @@ resources:
...
+DROP DATABASE {{ .PostgresDBName }}_migrated;
+DROP DATABASE {{ .PostgresDBName }}_backup;
```
katrybacka, line 295: [nit] another nit is that there is a command line tool for database deletion `dropdb` so you don't have to write SQLs; https://www.postgresql.org/docs/9.3/app-dropdb.html#:~:text=dropdb%20destroys%20an%20existing%20PostgreSQL,the%20SQL%20command%20DROP%20DATABASE.

katrybacka: It has `--if-exists` flag so it does not throw an error when the db does not exist

pkg/controller/command/command_config_bootstrap.go
```diff
@@ -266,3 +268,48 @@ resources:
...
+END_OF_SQL
+
+# Migrate old database dump to new one.
+commandutil migrate --in /backups/db.yml --out /backups/db_migrated.yml
+
+# Create a database for migrated data, initialize it with a new schema.
+createdb -w -h {{ .PostgresAddress }} -U {{ .PostgresUser }} {{ .PostgresDBName }}_migrated
+psql -v ON_ERROR_STOP=ON -w -h {{ .PostgresAddress }} -U {{ .PostgresUser }} -d {{ .PostgresDBName }}_migrated -f /usr/share/contrail/gen_init_psql.sql
+psql -v ON_ERROR_STOP=ON -w -h {{ .PostgresAddress }} -U {{ .PostgresUser }} -d {{ .PostgresDBName }}_migrated -f /usr/share/contrail/init_psql.sql
```
katrybacka, line 304: init_psql fills only id and ip pools, just a question isn't it handled by `commandutil convert`?

pkg/controller/command/command_controller.go
```diff
@@ -608,7 +602,13 @@ func (r *ReconcileCommand) listCommandsPods(commandName string) (*core.PodList,
 
 func (r *ReconcileCommand) prepareIntendedDeployment(instanceDeployment *apps.Deployment, commandCR *contrail.Command) error {
 	instanceDeploymentName := commandCR.Name + "-command-deployment"
-	intendedDeployment := contrail.SetDeploymentCommonConfiguration(instanceDeployment, &commandCR.Spec.CommonConfiguration)
+	podConfiguration := commandCR.Spec.CommonConfiguration
+	if commandCR.Status.UpgradeState == contrail.CommandShuttingDownBeforeUpgrade ||
```
katrybacka, line 606: if isUpgrading() didn't contain upgrade failed state, you could use it here :)

pkg/job/job.go
```diff
@@ -19,7 +19,7 @@ func (s Status) Pending() bool {
 	return true
 }
 
-func (s Status) Completed() bool {
+func (s Status) JobComplete() bool {
```
katrybacka, line 22: job is not necessary i think method's names should be completed and failed since the package and type `JobStatus` says what this adjective is about, but ofc if you would like to have `Job` in the method name, pls change also `Pending` method

katrybacka (changes requested):

pkg/controller/command/command_config_bootstrap.go
kaweue: As far as I understand it is not handled by `convert` however I think that those pools aren't in use any more.

kaweue: @cijohnson Could you confirm that ^^

pkg/job/job.go
kaweue: I cannot create method `Failed` since Status which is in fact batch.JobStatus already has this field.

I will change Pending to JobPending.

pkg/apis/contrail/v1alpha1/command_types.go
kaweue: changed

pkg/controller/command/command_config_bootstrap.go
kaweue: changed

pkg/controller/command/command_config_bootstrap.go
kaweue: done

pkg/controller/command/command_controller.go
kaweue: done

katrybacka (approved):

kaweue (approved):

PR #442, opened 2020-10-28
Add script for copying resource

Dysproz:
This is helper script which allow to copy resources between two clusters.
I've developed it as a help tool for side-by-side deployment of Openshift (where user has to manually copy CA secret abd csr signer configMap)

### DIFF
--- a/deploy/openshift/tools/resource-copier/README.md
+++ b/deploy/openshift/tools/resource-copier/README.md
@@ -1,0 +1,16 @@
+# RESOURCE COPIER
+
+## What is it?
+This script allows to copy one resource from one Kubernetes cluster to another (or from one namespace to another namespace in scope of the same cluster).
+
+## What do I need to use it?
+CLuster selection is based on kubeconfig file. So in order to access clusters, prepare kubeconfig file for both cluster from which you want to copy resource and cluster where you want to create copy of resource.
+
+## Usage
+"""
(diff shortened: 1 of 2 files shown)


Dysproz:
recheck ha

psykulsk (approved):

kaweue (approved):

deploy/openshift/tools/resource-copier/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+## What do I need to use it?
+CLuster selection is based on kubeconfig file. So in order to access clusters, prepare kubeconfig file for both cluster from which you want to copy resource and cluster where you want to create copy of resource.
+
+## Usage
+"""
+./resource-copier.sh --from-cluster <kubeconfig file of cluster with existing resource> --to-cluster <kubeconfig file of cluster where to copy resource> --from-namespace <target namespace where is original resource> --to-namespace <target namespace where to copy resource> --resource <resource type> --name <name of the resource>
+"""
+
+If one of kubeconfigs is not passed then script will automatically perform copy in scope of the same cluster.
```
katrybacka, line 14: what about a situation when you have multiple clusters defined in different contexts in one kubeconfig file?

katrybacka (changes requested):

deploy/openshift/tools/resource-copier/README.md
katrybacka: if it is not completely a case @Dysproz i will drop this comment but for me it really common to use one kubeconfig for multiple contexts

Dysproz: Not really in this case. If we'll need that feature then I'll add it for sure :)

katrybacka (approved):
ok understood

PR #450, opened 2020-11-03
Configure retry time for external keystone

michalskalski:
Add more control on how often operator will check availability
of external keystone endpoint.


michalskalski:
recheck

michalskalski:
recheck

katrybacka (approved):

kaweue:
Why there is a need to monitor external keystone?

pkg/controller/keystone/keystone_controller.go
```diff
@@ -145,7 +146,8 @@ func (r *ReconcileKeystone) Reconcile(request reconcile.Request) (reconcile.Resu
 		if ready {
 			return reconcile.Result{}, r.updateStatusWithExternalKeystone(keystone)
 		}
-		return reconcile.Result{}, err
+		reqLogger.Info(err.Error())
+		return reconcile.Result{RequeueAfter: time.Duration(keystone.Spec.ServiceConfiguration.ExternalAddressRetrySec) * time.Second}, nil
```
kaweue, line 150: Should we return err here?

kaweue (changes requested):

pkg/controller/keystone/keystone_controller.go
michalskalski: In case of cc, operator is deployed first and it waits until keystone will be deployed on openstack cluster. It will be expected behavior to not have keystone available to us on the beginning, that is why it is checked here and not treated as an error. Returning error could change the time after which next reconcile loop iteration happens, some resources claims it should trigger reconcile immediately on which previous approach based on, but from my observation it behaves differently.

kaweue (approved):
I don't like the fact that keystone controller is used to manage and perform health check or readiness check of external services. It should be only used to deploy and manage keystone deployment in k8s. Keystone address should be pass as parameter to services that use it, regardless if it is local one or external one.

Moreover operator should work only having access to k8s api, but now it has to have access to some external service. That might be problematic if we run it locally so not in the cluster.

However fixing that will require a lot of work, so I will approve this change.

michalskalski:
> I don't like the fact that keystone controller is used to manage and perform health check or readiness check of external services. It should be only used to deploy and manage keystone deployment in k8s. Keystone address should be pass as parameter to services that use it, regardless if it is local one or external one.

> Moreover operator should work only having access to k8s api, but now it has to have access to some external service. That might be problematic if we run it locally so not in the cluster.

There is an use case for using existing keystone, I understand it does not perfectly fit to native way of handling resource, but  does k8s resources always operates without need of accessing external resources?  Yes we are waiting for readiness of keystone, as other services waits for postgres or swift. That is not typical event based watch which is dictated of external nature of this resource. The services which uses keystone, get information like address from keystone resource status, I did not change that. I also reused already existing keystone client, with required changes introduced to it. Not sure if understand what you mean by running locally? The external keystone related code (including checking if it possible to fetch token) is only activate when the ExternalAddress property is set for keystone resource, it is not a default behavior, must be user decision.

> However fixing that will require a lot of work, so I will approve this change.
How would you address this use case differently? I can work further on this.

PR #454, opened 2020-11-09
Update openshift manifests to support ubi suffix for all images

Dysproz:
As from now on all images from this repository will be pushed with -ubi suffix then I update tags for Openshift manifests.
After this change user will have to provide full tag (together with suffix) to properly render manifests

### DIFF
--- a/deploy/openshift/manifests/00-contrail-09-manager.yaml
+++ b/deploy/openshift/manifests/00-contrail-09-manager.yaml
@@ -40,19 +40,19 @@ spec:
           analyticsStatisticsTTL: 2
           containers:
             - name: analyticsapi
-              image: <CONTRAIL_REGISTRY>/contrail-analytics-api:<CONTRAIL_VERSION>-ubi
+              image: <CONTRAIL_REGISTRY>/contrail-analytics-api:<CONTRAIL_VERSION>
             - name: api
-              image: <CONTRAIL_REGISTRY>/contrail-controller-config-api:<CONTRAIL_VERSION>-ubi
+              image: <CONTRAIL_REGISTRY>/contrail-controller-config-api:<CONTRAIL_VERSION>
             - name: collector
-              image: <CONTRAIL_REGISTRY>/contrail-analytics-collector:<CONTRAIL_VERSION>-ubi
+              image: <CONTRAIL_REGISTRY>/contrail-analytics-collector:<CONTRAIL_VERSION>
             - name: devicemanager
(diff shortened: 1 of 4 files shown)


deploy/openshift/manifests/00-contrail-09-manager.yaml
```diff
@@ -40,19 +40,19 @@ spec:
           analyticsStatisticsTTL: 2
           containers:
             - name: analyticsapi
-              image: <CONTRAIL_REGISTRY>/contrail-analytics-api:<CONTRAIL_VERSION>-ubi
+              image: <CONTRAIL_REGISTRY>/contrail-analytics-api:<CONTRAIL_VERSION>
```
kaweue, line 43: Why are you deleting UBI here?

kaweue: Ok I get it :)

kaweue (approved):

deploy/openshift/manifests/00-contrail-09-manager.yaml
Dysproz: Because now user will have to specify full tag in config file for script generating manifests. Before we had to take tag without suffix, because some images were with and some without suffix. Now it'll be more consistent :)

Dysproz: Although, these manifests are now used only for development. Customer should use release manifests already generated with release images

deploy/openshift/releases/R2011/manifests/00-contrail-09-manager.yaml
```diff
@@ -121,7 +121,7 @@ spec:
           - name: init
             image: python:3.8.2-alpine
           - name: provisioner
-            image: hub.juniper.net/contrail/contrail-operator-provisioner:2011.31
+            image: hub.juniper.net/contrail/contrail-operator-provisioner:2011.31-ubi
```
psykulsk, line 124: Could you change `hub.juniper.net/contrail` to `hub.juniper.net/contrail-nightly` in the release manifests as part of this PR? I think it would be better this way, as then we can actually tests those release manifests.

psykulsk (changes requested):

deploy/openshift/releases/R2011/manifests/00-contrail-09-manager.yaml
Dysproz: Oki doki!

katrybacka (approved):

psykulsk (approved):

PR #456, opened 2020-11-10
Command add support for external endpoints

kaweue:


pkg/controller/command/command_config_maps.go
```diff
@@ -1,15 +1,19 @@
...
 
+	"github.com/google/uuid"
 	corev1 "k8s.io/api/core/v1"
 
 	contrail "github.com/Juniper/contrail-operator/pkg/apis/contrail/v1alpha1"
 	"github.com/Juniper/contrail-operator/pkg/certificates"
 	"github.com/Juniper/contrail-operator/pkg/k8s"
 )
 
+var NameSpaceCommand = uuid.Must(uuid.Parse("362ad2d9-9460-4f65-a702-2cd99fbb0f64"))
```
katrybacka, line 15: i don't understand the name of this variable, what did you mean?

pkg/controller/command/command_config_maps.go
```diff
@@ -49,16 +53,31 @@ func (c *configMaps) ensureCommandConfigExist(postgresAddress, ConfigEndpoint st
...
+		{Name: "nodejs", PrivateURL: webUIURL, PublicURL: webUIURL},
+		{Name: "telemetry", PrivateURL: telemetryURL, PublicURL: telemetryURL},
+		{Name: "config", PrivateURL: configAPIURL, PublicURL: configAPIURL},
+		{Name: "keystone", PrivateURL: keystoneURL, PublicURL: keystoneURL},
+		{Name: "swift", PrivateURL: swiftProxyURL, PublicURL: swiftProxyURL},
+	}
+	ces = append(ces, c.ccSpec.ServiceConfiguration.Endpoints...)
+	bes := []bootstrapEndpoint{}
+	for _, e := range ces {
+		bes = append(bes, *newBootstrapEndpoint(e))
```
katrybacka, line 72: do we need to return an pointer from newBootstrapEndpoint method if you are just using objects and then you append the object to the slice

katrybacka (changes requested):

lrebeja:
Maybe we can allow user to configure this port as well?

kaweue:
@kaweue Which port?

pkg/controller/command/command_config_maps.go
kaweue: Added a comment

pkg/controller/command/command_config_maps.go
kaweue: changed

kabart (approved):

katrybacka (approved):

PR #457, opened 2020-11-12
Create certs for all running pods

kaweue:


katrybacka (approved):

pkg/apis/contrail/v1alpha1/base_types.go
```diff
@@ -314,6 +314,9 @@ func PodsCertSubjects(podList *corev1.PodList, hostNetwork *bool, podAltIPs PodA
 		useNodeName = *hostNetwork
 	}
 	for _, pod := range podList.Items {
+		if pod.Status.Phase == corev1.PodFailed {
+			continue
+		}
```
psykulsk, line 319: nit: why is this check necessary? What if containers of pod with PodFailed status will be restarted? Shouldn't they have correctly filled secrets mounted to them?

psykulsk (approved):

pkg/apis/contrail/v1alpha1/base_types.go
kaweue: I'm not sure if that is possible since
`   // PodFailed means that all containers in the pod have terminated, and at least one container has
    // terminated in a failure (exited with a non-zero exit code or was stopped by the system).`

Is it that a pod needs to be recreated?

psykulsk: Makes sense, thanks 👍

PR #464, opened 2020-11-18
Openshift 4.6 support

Dysproz:
This PR fix overlay mounts for CoreOS in Openshift 4.6.
Currently with this fix nodes are created and joined into cluster which enables installation of Openshift with Contrail.
However, installation cannot report complete status as there's a problem in machineconfigpool resource for master.
Openshift by deault consumes all machineconfigs for both master and worker nodes and then create another machine config resource which is a rendered ignition config for CoreOS used in final node boot.
MachineConfigPool resources lookup for these rendered configs and then use some internal logic to define whether created node is properly configured.
However, currently machineconfigpool for master nodes report that nodes are not configured properly as it cannot find rendered machineconfig for master nodes. It is so, because it tries to lookup rendered config with uuid number which is not the same as uuid of actually rendered machineconfig.

It may be debugged by exporting kubeconfig and running `kubectl get machineconfig -A`.
This command outputs all machineconfigs in the cluster and at the end there are machine configs called `master-rendered-<uuid>` which is rendered master config for nodes.
Looking into another command `kubectl get machineconfigpool -A` you may notice that machineconfigpool master has not ready state which is the reason why installation is not completed due to install binary, but the cluster works.


Dysproz:
recheck

PR #468, opened 2020-11-20
Change clientset to use config from config.GetConfig()

Janek3d:
During debugging in VS Code on cluster deployed on AWS, I was getting error that GetClientset function could not locate kubeconfig file despite providing "kubeconfig" flag and "KUBECONFIG" variable in launch.json file. After some digging I found that GetClientset uses getClientConfig function which was simply ignoring this flag and variable. I decided to get clientsert by using more direct function GetClientsetFromConfig and config gathered by config.GetConfig. I also deleted GetConfig function which was not used anywhere.


Dysproz (approved):
Great input Janusz!

cmd/manager/main.go
```diff
@@ -91,7 +91,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	clientset, err := v1alpha1.GetClientset()
+	clientset, err := v1alpha1.GetClientsetFromConfig(cfg)
```
kaweue, line 94: In my opinion you can also delete `GetClientsetFromConfig` and move `clientset, err := kubernetes.NewForConfig(config)` here. I really don't see a point of having a function that only calls one other function.

kaweue (changes requested):

cmd/manager/main.go
Janek3d: I changed it to `kubernetes.NewForConfig(cfg)`.

kaweue (approved):

PR #470, opened 2020-11-24
Add Service Type to Swift Proxy configuration.

kabart:


pkg/controller/swiftproxy/swiftproxy_controller.go
```diff
@@ -134,7 +134,7 @@ func (r *ReconcileSwiftProxy) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	servicePortsMap := map[int32]string{int32(swiftProxy.Spec.ServiceConfiguration.ListenPort): ""}
-	svc := r.kubernetes.Service(request.Name+"-swiftproxy", core.ServiceTypeLoadBalancer, servicePortsMap, contrail.SwiftProxyInstanceType, swiftProxy)
+	svc := r.kubernetes.Service(request.Name+"-swiftproxy", swiftProxy.Spec.ServiceConfiguration.SwiftServiceType, servicePortsMap, contrail.SwiftProxyInstanceType, swiftProxy)
```
kaweue, line 137: I think that we need to start with the default value set to core.ServiceTypeLoadBalancer. When SwiftServiceType is empty string we should use ServiceTypeLoadBalancer. If we pass empty string here we will get ClusterIP service. That might have an effect on other deployments.

kaweue (changes requested):
Please remove WIP and change the title of PR to something that relates to the change that you are introducing.

kaweue (approved):

pkg/apis/contrail/v1alpha1/swiftproxy_types.go
```diff
@@ -17,14 +17,15 @@ type SwiftProxySpec struct {
...
-	Containers            []*Container `json:"containers,omitempty"`
+	MemcachedInstance     string             `json:"memcachedInstance,omitempty"`
+	ListenPort            int                `json:"listenPort,omitempty"`
+	KeystoneInstance      string             `json:"keystoneInstance,omitempty"`
+	KeystoneSecretName    string             `json:"keystoneSecretName,omitempty"`
+	CredentialsSecretName string             `json:"credentialsSecretName,omitempty"`
+	SwiftConfSecretName   string             `json:"swiftConfSecretName,omitempty"`
+	RingConfigMapName     string             `json:"ringConfigMapName,omitempty"`
+	Containers            []*Container       `json:"containers,omitempty"`
+	SwiftServiceType      corev1.ServiceType `json:"swiftServiceType,omitempty"`
```
katrybacka, line 28: It would be nice to allow picking only valid service type and "", not any string, so then the validation of service type will be at api level, since right now i see that we can type in CR that the servicetype is "abc" and then the load balancer will be created. using kubebuilder you can create an enum which can have one value from the set e.g.: ClusterIP, NodePort, LoadBalancer, ExternalName, "". I think that would be a good feature

katrybacka (changes requested):

pkg/apis/contrail/v1alpha1/swiftproxy_types.go
```diff
@@ -25,10 +28,28 @@ type SwiftProxyConfiguration struct {
 	SwiftConfSecretName   string       `json:"swiftConfSecretName,omitempty"`
 	RingConfigMapName     string       `json:"ringConfigMapName,omitempty"`
 	Containers            []*Container `json:"containers,omitempty"`
+	SwiftServiceType      serviceType  `json:"swiftServiceType,omitempty"`
 	// Service name registered in Keystone, default "swift"
 	SwiftServiceName string `json:"swiftServiceName,omitempty"`
 }
 
+// GetServiceType returns Swift Proxy's service type.
+func (configuration SwiftProxyConfiguration) GetServiceType() corev1.ServiceType {
```
katrybacka, line 37: here you don't have to type this switch since the values of servType will be correct, you can just cast it to corev1.ServiceType

katrybacka (changes requested):

katrybacka (approved):

PR #474, opened 2020-11-25
Add keystone authentication mode to kubemanager's config

Janek3d:


pkg/apis/contrail/v1alpha1/kubemanager_types.go
```diff
@@ -82,6 +83,8 @@ type KubemanagerConfiguration struct {
 	RabbitmqUser          string       `json:"rabbitmqUser,omitempty"`
 	RabbitmqPassword      string       `json:"rabbitmqPassword,omitempty"`
 	RabbitmqVhost         string       `json:"rabbitmqVhost,omitempty"`
+	KeystoneSecretName    string       `json:"keystoneSecretName,omitempty"`
+	KeystoneInstance      string       `json:"keystoneInstance,omitempty"`
```
kaweue, line 87: There shouldn't be direct dependency on keystone CR in kubemanager controller and crd

kaweue: Keystone endpoint address should be a part of KubemanagerNodesConfiguration structure, like config, zookeeper ....

pkg/apis/contrail/v1alpha1/kubemanager_types.go
```diff
@@ -184,7 +187,13 @@ func (c *Kubemanager) InstanceConfiguration(request reconcile.Request,
 		}
 		kubemanagerConfig.ServiceSubnets = serviceSubnets
 	}
-
+	keystoneAuth := &KubemanagerAuthParameters{}
+	if configNodesInformation.AuthMode == AuthenticationModeKeystone {
+		err := c.AuthParameters(keystoneAuth, client)
```
kaweue, line 192: Authentication parameters should be injected from manager.go 

Here we should for keystone to become active https://github.com/Juniper/contrail-operator/blob/ec8d40e3323efdf7b6f89a12957d7479b419bcbd/pkg/controller/manager/manager_controller.go#L571

and here we should create a configuration https://github.com/Juniper/contrail-operator/blob/ec8d40e3323efdf7b6f89a12957d7479b419bcbd/pkg/controller/manager/manager_controller.go#L1066

kaweue (changes requested):

pkg/apis/contrail/v1alpha1/kubemanager_types.go
katrybacka: +1

pkg/apis/contrail/v1alpha1/kubemanager_types.go
```diff
@@ -62,26 +62,27 @@ type KubemanagerServiceConfiguration struct {
...
+	KubernetesClusterName string             `json:"kubernetesClusterName,omitempty"`
+	IPFabricSubnets       string             `json:"ipFabricSubnets,omitempty"`
+	IPFabricForwarding    *bool              `json:"ipFabricForwarding,omitempty"`
+	IPFabricSnat          *bool              `json:"ipFabricSnat,omitempty"`
+	KubernetesTokenFile   string             `json:"kubernetesTokenFile,omitempty"`
+	HostNetworkService    *bool              `json:"hostNetworkService,omitempty"`
+	RabbitmqUser          string             `json:"rabbitmqUser,omitempty"`
+	RabbitmqPassword      string             `json:"rabbitmqPassword,omitempty"`
+	RabbitmqVhost         string             `json:"rabbitmqVhost,omitempty"`
+	AuthMode              AuthenticationMode `json:"authMode,omitempty"`
```
Dysproz, line 85: Is it possible to make in enum that only noauth and keystone values will be accepted?

pkg/apis/contrail/v1alpha1/templates/kubemanager_config.go
```diff
@@ -61,11 +61,12 @@ use_ssl = True
...
-;AUTHN_PORT = 35357
-;AUTHN_URL = /v2.0/tokens
+AUTHN_TYPE = {{ .AuthMode }}
+AUTHN_PROTOCOL = {{ .KeystoneAuthProtocol }}
+AUTHN_SERVER = {{ .KeystoneAddress }}
+AUTHN_PORT = {{ .KeystonePort }}
+AUTHN_DOMAIN = {{ .KeystoneUserDomainName }}
+cafile = {{ .CAFilePath }}
+AUTHN_URL = /v3/auth/tokens
 ;AUTHN_TOKEN_URL = http://127.0.0.1:35357/v2.0/tokens
```
psykulsk, line 71: Maybe this line could be removed as it is commented anyway?

pkg/apis/contrail/v1alpha1/kubemanager_types.go
```diff
@@ -454,6 +467,10 @@ func (c *Kubemanager) ConfigurationParameters() KubemanagerConfiguration {
 		ipFabricSnat = KubernetesIPFabricSnat
 	}
 
+	if c.Spec.ServiceConfiguration.AuthMode != AuthenticationModeNoAuth && c.Spec.ServiceConfiguration.AuthMode != AuthenticationModeKeystone {
```
Dysproz, line 470: I don't like it that we check every available mode. What if we'll add more auth modes in the future? That'll make it a pretty long statement. Cannot it be checked somehow different? Maybe enum on struct would ensure that passed vale makes sense or simple check that it's not empty.

Dysproz (changes requested):

psykulsk (changes requested):

pkg/apis/contrail/v1alpha1/base_types.go
```diff
@@ -978,6 +978,29 @@ func NewConfigClusterConfiguration(name string, namespace string, myclient clien
...
+
+	keystoneConfig := keystoneInstance.ConfigurationParameters()
+	endpoint := keystoneInstance.Status.Endpoint + ":" + strconv.Itoa(keystoneConfig.ListenPort)
+	keystoneCluster = KeystoneClusterConfiguration{
+		Endpoint:       endpoint,
+		Port:           keystoneConfig.ListenPort,
+		AuthProtocol:   keystoneConfig.AuthProtocol,
+		UserDomainName: keystoneConfig.UserDomainName,
+	}
+	return keystoneCluster, nil
```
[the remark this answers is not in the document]
Janek3d: In the same way it is created for Cassandra's cluster https://github.com/Juniper/contrail-operator/blob/master/pkg/apis/contrail/v1alpha1/base_types.go#L832-L839 so I thought that it should be this way, will change it.

deploy/crds/contrail.juniper.net_kubemanagers_crd.yaml
```diff
@@ -195,6 +200,24 @@ spec:
...
+                  description: KeystoneClusterConfiguration defines all information
+                    about Keystone's endpoints.
+                  properties:
+                    address:
+                      type: string
+                    authMode:
+                      enum:
+                      - noauth
+                      - keystone
+                      type: string
```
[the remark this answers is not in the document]
Janek3d: Deleted it

pkg/apis/contrail/v1alpha1/kubemanager_types.go
Janek3d: Changed it to only check if AuthMode is empty string.

pkg/apis/contrail/v1alpha1/base_types.go
```diff
@@ -1117,3 +1140,26 @@ type VrouterClusterConfiguration struct {
 	Gateway           string
 	MetaDataSecret    string
 }
+
+// KeystoneClusterConfiguration defines all information about Keystone's endpoints.
+type KeystoneClusterConfiguration struct {
+	AuthMode       AuthenticationMode `json:"authMode,omitempty"`
+	Endpoint       string             `json:"address,omitempty"`
```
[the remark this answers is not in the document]
Janek3d: Changed

kaweue (approved):

Dysproz (approved):

psykulsk (approved):

PR #476, opened 2020-11-26
Add LoadBalancerIP to Swift Proxy's certificate as an alternative IP

kabart:


pkg/controller/swiftproxy/swiftproxy_controller.go
```diff
@@ -152,7 +152,11 @@ func (r *ReconcileSwiftProxy) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, fmt.Errorf("failed to list swift proxy pods: %v", err)
 	}
 
-	if err := r.ensureCertificatesExist(swiftProxy, swiftProxyPods, swiftProxy.Status.ClusterIP); err != nil {
+	var alternativeIP []string
+	if swiftProxy.Status.LoadBalancerIP != "" {
+		alternativeIP = append(alternativeIP, swiftProxy.Status.LoadBalancerIP)
+	}
+	if err := r.ensureCertificatesExist(swiftProxy, swiftProxyPods, swiftProxy.Status.ClusterIP, alternativeIP...); err != nil {
```
kaweue, line 159: LoadBalancerIP may be assigned with some delay or even may not be assigned at all. 
In case of significant delay I see potential problem.
1) LoadBalancerIP is empty
2) Certificate is created
3) SwiftProxy is started and reads certificate without LoadBalancerIP
4) LoadBalancerIP is set
5) Certificate is updated
6) SwiftProxy may be still using certificate without LoadBalancerIP. I'm not sure if it can reload those files automatically.

Could you please check that?
If that is the case swiftproxy controller needs to wait until LoadBalancerIP is set or service needs to be reloaded

kaweue (changes requested):

kabart:
recheck ha

kabart:
recheck ha

kabart:
recheck ha

pkg/apis/contrail/v1alpha1/swiftproxy_types.go
```diff
@@ -68,17 +68,17 @@ type SwiftProxyList struct {
...
 	return PodsCertSubjects(podList, s.Spec.CommonConfiguration.HostNetwork, altIPs)
 }
 
-// GetServiceType returns chosen Service type for Swift Proxy, default is LoadBalancer
+// GetServiceType returns chosen Service type for Swift Proxy, default is ClusterIP
 func (s *SwiftProxy) GetServiceType() corev1.ServiceType {
 	st := s.Spec.ServiceConfiguration.Service.GetServiceType()
 	if st == "" {
-		return corev1.ServiceTypeLoadBalancer
+		return corev1.ServiceTypeClusterIP
```
kaweue, line 81: LoadBalancer service should be the default value. The default ServiceTypeClusterIP will break other projects

kaweue (changes requested):

kabart:
recheck ha

pkg/apis/contrail/v1alpha1/swiftproxy_types.go
kabart: Corrected.

katrybacka (approved):

kaweue (approved):

PR #477, opened 2020-11-26
Add annotation to Swift Proxy LoadBalancer Service to pick correct MetalLB IP pool

katrybacka:


pkg/apis/contrail/v1alpha1/swiftproxy_types.go
```diff
@@ -29,17 +29,12 @@ type SwiftProxyConfiguration struct {
 	RingConfigMapName     string       `json:"ringConfigMapName,omitempty"`
 	Containers            []*Container `json:"containers,omitempty"`
 	SwiftServiceType      serviceType  `json:"swiftServiceType,omitempty"`
+	// MetalLB pool name that should be used by Service
+	IPPoolName string `json:"ipPoolName,omitempty"`
```
kaweue, line 33: Is it too specific? That we create metalb field in this api? Maybe we should create some sub structure like
```
type Service struct {
type 
annotations
}
```
That might be more general not limited to metallb

katrybacka: so you want to pass annotations specified by the user?

pkg/apis/contrail/v1alpha1/swiftproxy_types.go
```diff
@@ -78,12 +73,32 @@ type SwiftProxyList struct {
...
+// GetServiceAnnotations prepares annotations for LoadBalancer Service
+func (s *SwiftProxy) GetServiceAnnotations() map[string]string {
+	if s.GetServiceType() == corev1.ServiceTypeLoadBalancer {
+		return map[string]string{}
+	}
+	poolName := "mgmt"
+	if s.Spec.ServiceConfiguration.IPPoolName != "" {
+		poolName = s.Spec.ServiceConfiguration.IPPoolName
+	}
+	return map[string]string{"metallb.universe.tf/address-pool": poolName}
```
kaweue, line 99: Please add TODO to remove this default value

katrybacka: done

pkg/apis/contrail/v1alpha1/swiftproxy_types.go
katrybacka: done

kaweue (approved):

kabart (approved):

PR #478, opened 2020-11-26
Modified images pulled from docker.io to be pulled from hub.juniper.net

psykulsk:


psykulsk:
recheck ha

Dysproz (approved):
Very good change Piotr!

Janek3d (approved):

PR #486, opened 2020-12-02
Update operator to v0.18.2

kaweue:
Updating operator to newer version 0.18.2 however without updating crds versions to v1 it will require small refactor. I hope to do it in next PRs.


michalskalski:
cool, that would mean that we can start use `kubebuilder:default` to set default values.

kaweue:
> cool, that would mean that we can start use `kubebuilder:default` to set default values.

Not yet, but I hope that soon it will be possible. 
In crds in version v1 there is problem with embedded metadata structures. Basically they are ignored and not stored at kubernetes API server. It is connected to https://kubernetes.io/blog/2019/06/20/crd-structural-schema/#pruning-don-t-preserve-unknown-fields and to the fact that properties for metadata struct is not generated for some reason.

kaweue:
recheck

kaweue:
recheck

Dysproz (approved):

michalskalski (approved):

katrybacka (approved):

PR #487, opened 2020-12-02
Make swift deployment optional

michalskalski:
Swift is used to support one feature of contrail command, which is not necessary always used, but requires whole swift stack.
Deploy command also when SwiftInstance is not defined.

### DIFF
--- a/pkg/controller/command/command_controller_test.go
+++ b/pkg/controller/command/command_controller_test.go
@@ -43,7 +43,7 @@ func TestCommand(t *testing.T) {
 		},
 		"Swift secret name is empty": {
 			initObjs: []runtime.Object{
-				newCommand(true),
+				newCommand(),
 				newConfig(true),
 				newPostgres(true),
 				newAdminSecret(),
@@ -78,7 +78,7 @@ func TestCommand(t *testing.T) {
 	}{
 		"no Postgres": {
 			initObjs: []runtime.Object{
-				newCommand(true),
+				newCommand(),
 				newConfig(true),
 				newAdminSecret(),
 				newSwiftSecret(),
@@ -90,7 +90,7 @@ func TestCommand(t *testing.T) {
 		},
 		"no Swift": {
 			initObjs: []runtime.Object{
-				newCommand(true),
+				newCommand(),
 				newConfig(true),
 				newPostgres(true),
 				newAdminSecret(),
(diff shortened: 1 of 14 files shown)


michalskalski:
recheck ha

michalskalski:
recheck ha

michalskalski:
recheck ha

michalskalski:
recheck

michalskalski:
recheck ha

pkg/controller/command/command_controller_test.go
```diff
@@ -772,8 +791,12 @@ func newCertSecret(ips []string) *core.Secret {
 	}
 }
 
-func newCommand() *contrail.Command {
+func newCommand(withSwift bool) *contrail.Command {
```
kaweue, line 794: I would prefer a new method something like:
newCommandWithSwift or newCommandWithoutSwift

pkg/controller/command/command_controller_test.go
```diff
@@ -1200,10 +1223,15 @@ func assertConfigMap(t *testing.T, actual *core.ConfigMap) {
 		},
 	}, actual.ObjectMeta)
 
-	assert.Equal(t, expectedCommandConfig, actual.Data["command-app-server0.0.0.0.yml"])
+	expectedConfig := expectedCommandConfig
+	if *swift == "" {
```
katrybacka, line 1227: i don't think that pointer is necessary here, `swift` could be just a string

pkg/controller/command/command_controller_test.go
```diff
@@ -1415,6 +1447,13 @@ replication:
     enabled: false
 `
 
+const absentCommandWithNoSwift = `
```
katrybacka, line 1450: maybe it could be named `swiftUserConfiguration` i think it's more accurate

pkg/controller/command/command_controller_test.go
```diff
@@ -1637,3 +1676,17 @@ resources:
       public_url: https://40.40.40.40:5080
     kind: endpoint
 `
+const absentInitWithNoSwift = `  - data:
```
katrybacka, line 1679: and here it could be called `swiftEndpoint`

katrybacka (changes requested):

kaweue (approved):

katrybacka (approved):

PR #488, opened 2020-12-02
Reduce Patroni priviliges in k8s cluster

katrybacka:


katrybacka:
recheck ha

katrybacka:
recheck ha

pkg/controller/postgres/postgres_controller.go
```diff
@@ -272,6 +274,119 @@ func (r *ReconcilePostgres) ensureServicesExist(postgres *contrail.Postgres) (le
...
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: clusterRoleBindingName}, existingClusterRoleBinding)
+	if errors.IsNotFound(err) {
+		clusterRoleBinding := &rbac.ClusterRoleBinding{
+			TypeMeta: meta.TypeMeta{
+				APIVersion: "rbac/v1",
+				Kind:       "ClusterRoleBinding",
+			},
+			ObjectMeta: meta.ObjectMeta{
+				Name: clusterRoleBindingName,
+			},
```
psykulsk, line 370: Shouldn't there be a namespace here? Because now the permissions are granted for objects across all namespaces. Is that intentional?

katrybacka: No, no, clusterrole is not namespace-scoped, the subject is namespace-scoped

psykulsk: Hmm, I found this example in k8s docs and it looks like the binding has to be namespace-scoped for the permissions to be granted only in that namespace. Have a look. But perhaps I'm wrong.

https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-example
```
apiVersion: rbac.authorization.k8s.io/v1
# This role binding allows "dave" to read secrets in the "development" namespace.
# You need to already have a ClusterRole named "secret-reader".
kind: RoleBinding
metadata:
  name: read-secrets
  #
  # The namespace of the RoleBinding determines where the permissions are granted.
  # This only grants permissions within the "development" namespace.
  namespace: development
subjects:
- kind: User
  name: dave # Name is case sensitive
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: secret-reader
  apiGroup: rbac.authorization.k8s.io
```

katrybacka: but it is a rolebinding what you send and  it is namespace-scoped, I create CLUSTERROLEBINDING and it is not namespace-scoped

psykulsk: Right, my bad, sorry :D

psykulsk (approved):

kaweue (approved):

PR #498, opened 2020-12-04
Migrate crds to apiextensions.k8s.io/v1

kaweue:
Migration crds versions to v1.

crds in version v1 doesn't handle embedded  metav1.ObjectMeta fields. That is why a wrapper on metav1.ObjectMeta was created in package contrail. The consequence is that manager structure doesn't embed other resources. Instead of that a Service types were created with only desired metadata and spec (without status)


kaweue:
recheck ha

kaweue:
recheck ha

katrybacka (approved):

Dysproz (approved):

PR #500, opened 2020-12-07
Check if no. of failed pods exceeded BackOffLimit

katrybacka:
I found that there is a bug in k8s that job-controller responsible for setting Failed Condition on Jobs does not work correctly. Sometimes the Status is not updated or updated with a huge delay. As the result multiple pods are ran and the number of them exceeds the BackOffLimit specified in Job.Spec. 
This bug sometimes causes the failure of nightly job.
This PR contains a workaround that Job is assumed as Failed when it has Failed Condition set to true or the number of failed Pods i equal or greater than BackOffLimit.


katrybacka:
recheck ha

Dysproz (approved):

kaweue (approved):

PR #516, opened 2020-12-16
In CI use local registry name instead of ip address.

kaweue:
Since we are moving to multiple node test environment, a single registry for all nodes will introduce latency.
That is why every worker node will have its own docker registry proxy running as a docker container. Tests scripts will use its name to get its IP inside "kind" docker network, which is used to spawn test clusters.


kaweue:
recheck ha

psykulsk (approved):

katrybacka (approved):

PR #517, opened 2020-12-16
Improve data IP discover

michalskalski:
The old way failed to get IP from data subnet if interface was
configured with Network Manager which adds additional fields after
src IP. In new way the string just after 'src' will be taken into
account.

### DIFF
--- a/pkg/controller/utils/utils.go
+++ b/pkg/controller/utils/utils.go
@@ -671,7 +671,7 @@ func getPodInitStatus(reconcileClient client.Client,
 
 							if cidr, ok := pod.Annotations["dataSubnet"]; ok {
 								if cidr != "" {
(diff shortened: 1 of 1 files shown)


michalskalski:
recheck ha

katrybacka:
instead of merging a separate PR, it should be done by merging R2011 to master (the change is already o R2011)

michalskalski:
> instead of merging a separate PR, it should be done by merging R2011 to master (the change is already o R2011)

Yes that is correct, although there is a difference between 2 branches where this part of the code is located and on the end I would need rewrite 2011 commit to this form.

katrybacka:
I understand but this is not a complex commit so the conflict should not be that hard to fix, am I right? In that case, the R2011 will remain easy to merge to the master. Having two different commits on both branches will make it harder :<

michalskalski:
Ok I will try merge 2011 to master.

PR #537, opened 2021-04-12
adding node_timeout to 90 sec, it was missing  in swift proxy, openstack has default value of 10.

vijayrap:
adding node_timeout to 90 sec, inwas missing  in swift proxy


vijayrap:
check HA

cijohnson (approved):

